PR #329, opened 2014-12-14
Fixed most hairlines for iPhone6 plus and other retina devices

MarcGodard:
Before and after screenshots below.

Hairlines are more consistent throughout and also now visible in the browser (chrome when you set iPhone6 plus emulation).

TODO: Figure out how to improve hairlines for input boxes (messagebar and prompt) without impacting style or performance.

screen shot 2014-12-14 at 11 14 16 am
screen shot 2014-12-14 at 11 15 12 am

 

ios simulator screen shot dec 14 2014 11 09 35 am
ios simulator screen shot dec 14 2014 11 04 02 am
ios simulator screen shot dec 14 2014 11 09 45 am
ios simulator screen shot dec 14 2014 11 05 03 am

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 build
 custom
 npm-debug.log
+.idea
--- a/src/less/content-block.less
+++ b/src/less/content-block.less
@@ -31,10 +31,13 @@
     border-bottom: 1px solid @contentBlockBorderColor;
     border-top: 1px solid @contentBlockBorderColor;
     color: #000;
-    html.retina.ios-gt-7 & {
-        border-top-width: 0.5px;
-        border-bottom-width: 0.5px;
-    }
+}
+html.retina.ios-gt-7 .content-block-inner {
+    position: relative;
+    border-top-width: 0;
+    border-bottom-width: 0;
+    .hairline-top();
+    .hairline-bottom();
 }
 .content-block.inset {
     margin-left: 15px;
--- a/src/less/lists.less
+++ b/src/less/lists.less
@@ -395,6 +395,7 @@
 html.retina.ios-gt-7 {
     .list-block {
         ul {
+            position: relative;
             border-top-width: 0;
             border-bottom-width: 0;
             .hairline-top();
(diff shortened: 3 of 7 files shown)


MarcGodard:
@nolimits4web Just wondering if you had looked at this yet?

I think its good work, but maybe you can improve it.

Also, side note, I added (again in my testing branch) an infinite scroll for the top of the page. This is mostly for the messaging so you can load more (older) messages.

The only issue is you cannot set the event listener until after the messages are loaded, if not it loads more on top and scroll from the bottom of the messages.

johnjamesjacoby:
Confirming this is an issue on the iPhone 6+. `0.5px` gets scaled down to `0` as `1/3` is less than `½`.

nolimits4web:
Yes, i had looked. And as @johnjamesjacoby mentioned, i am trying to look for a better approach. The biggest problem of such approach is the later custom customisation of such borders. I am sure that there will be a lot of troubles and questions from dev.

I haven't real iPhone 6 Plus device, so i ask you to check it on real device. From what i have tried in simulator:
- setting CSS border-width to 0.7px gives "stable" (not buggy) 2 physical-pixels border width (instead of 3px like with 1px border). This is already not bad
- setting CSS border-width to 0.67px gives 1 physical-pixels border width. But for some reason, it becomes buggy and such borders disappear during scrolling. And i don't know, maybe it is just a simulator bug. So if it is fine on real iPhone 6 Plus then it could be a "holy grail" solution :)

nolimits4web:
@MarcGodard i saw your infinite scroll on top. I think this is could be useful, so i will add that code to F7

johnjamesjacoby:
Here's how I've circumvented it for now: https://github.com/nolimits4web/Framework7/commit/8ed43b470f79fddb02eeafbaf52adb071fafd2a1

Then in all the LESS files, instead of targeting `.retina` target `.pixel-ratio-2`. In my experience, a `1px` border on the iPhone 6+ isn't horrible, thanks to the immense amount of pixels and the overall scaling that takes place to fit everything.

MarcGodard:
I agree that my approach adds complications. As far as left and right borders, I agree, this is still a problem.

The benefits of this approach is that it will be visible on all devices and browsers as scaling down cannot go to 0 and will go to the smallest visible line.

@johnjamesjacoby I prefer the lines to be one visible pixel (3px in css). The thicker lines don't look as nice on the iphone 6 plus.

johnjamesjacoby:
Yes. It doesn't appear to work for me. Anything up to `0.9px` still had the same rounding-down effect of being invisible.

MarcGodard:
@nolimits4web When I look at the changed files for this pull request, all the correct changes are there. There shouldn't be anything about the lazy load.

These changes were done on my master branch as I didn't quite know what i was doing when i started contributing.

I played around with 0.33 scaling and it also works well, tested on iPhone 6plus, and retina iPad mini.

nolimits4web:
Just merged #350 based on @MarcGodard pseudo selectors solution

PR #330, opened 2014-12-14
Ionic Keyboard Plugin with cordova

MarcGodard:
Added an option that the ionic keyboard cordova plugin is installed, which when using the message bar will add the correct space below the messages, scroll to position, and then open the keyboard with the message bar on top.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 build
 custom
 npm-debug.log
+.idea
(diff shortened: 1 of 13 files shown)


nolimits4web:
Just back from vacation, will check your improvements during a week

MarcGodard:
Sounds good. I think the keyboard one could be improved more.

PR #334, opened 2014-12-19
Fix navbar manipulation without animation

Sl1v3r:
Continue of [#322](https://github.com/nolimits4web/Framework7/pull/322) 
Now it breaks with animatePages: false
updated fiddle: http://jsfiddle.net/Sl1v3r/h6ycoyps/

PS: This fiddle throws Uncaught TypeError, it seems to be another bug, just run it few times and when it is loaded without errors, randomly click digits like you did before.


nolimits4web:
Merged, thanks

PR #353, opened 2015-01-16
Infinite Scroll Top

MarcGodard:
After you scroll to the bottom of a page, initiate the listener for the
event, when you scroll up to the top, load more content. This is best
to go with the messaging portion of F7.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 build
 custom
 npm-debug.log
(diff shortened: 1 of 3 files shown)


MarcGodard:
infinittop

nolimits4web:
Thanks, added with much simpler way, i guess no need to use here fully separated function/methods

PR #355, opened 2015-01-16
Allows loading of old messages

MarcGodard:
Used with the infinite top scrolling this can load images from the top
of the messages.

Can also change the method if using a different library or can create a custom one.

infinittop

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 build
 custom
 npm-debug.log
+.idea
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, but we still need option to have contra-versa logic with new messages on top/first, i will fix now

nolimits4web:
Add some improvements to keep option with new messages on top, also fixed animation a bit, so the animation for top messages will be from top

MarcGodard:
Very Nice.

MarcGodard:
The only problem I see is that people will need to add and remove the class new-messages-first depending on if infinite scrolling is loading more or if they are typing and adding a new message on the bottom.

nolimits4web:
Why? It is just adds another contra versa permanent scenario when you have new messages on top, old on bottom, and use of infinite scroll on bottom to load older messages

MarcGodard:
Maybe, let me implement it and I will see.

PR #357, opened 2015-01-16
Prevent scrolling through overlays

MarcGodard:
make the touch move event function like a click event to prevent
scrolling through the overlay

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 build
 custom
 npm-debug.log
+.idea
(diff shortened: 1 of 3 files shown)


nolimits4web:
Was thinking that there could be solution without JS, but seems to be not. So i think there is a better place for it, for example in 'clicks.js' when we may prevent it once on all overlays, like:
```
$(document).on('touchstart', '.panel-overlay, .modal-overlay, .preloader-indicator-overlay, .popup-overlay, .searchbar-overlay', function (e) {
    e.preventDefault();
});

PR #358, opened 2015-01-16
Allows padding to be added by a keyboard

MarcGodard:
Instead of losing the screen when the keyboard opens, sometimes you
want to move the content of the page.

This allows you to use a plugin to do so by allowing you to specify the
current keyboard height.

Also, initial padding can now be different (like if you modify the
message bar).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 build
 custom
 npm-debug.log
+.idea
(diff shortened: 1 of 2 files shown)


MarcGodard:
Done.

nolimits4web:
Merged, thanks

PR #366, opened 2015-01-22
Added external click for cordova

MarcGodard:
I tried many ways to not include this in the library, but it must be.

The _system target is original enough and should only ever be used for cordova apps.


MarcGodard:
This is ready to merge :)

nolimits4web:
Merged, thanks

PR #367, opened 2015-01-23
use new class smart-select-value to set selected option in Smart Select

yysun:
Smart Select automatically sets div.item-after to be the selected option. But there is another use case missing that is to set selected option based on a value. This change is to implement the missing use case. 

Adding a value to div.item-after and also add a new class smart-select-value to it will set the select option which matches the value to be selected.


nolimits4web:
Merged, thanks

PR #368, opened 2015-01-23
Update lists.less

fouber:
in ``position: absolute`` mode, 

```css
.sortable-handler {
    position: absolute;
    top: 0;
    height: ~"-webkit-calc(100% - 1px)";
    height: ~"calc(100% - 1px)";
}
```

is equivalent to

```css
.sortable-handler {
    position: absolute;
    top: 0;
    bottom: 1px;
}
```

and can produce better compatibility

### DIFF
--- a/src/less/lists.less
+++ b/src/less/lists.less
@@ -334,8 +334,7 @@
         position: absolute;
         right: 0;
         top: 0;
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged,thanks

PR #374, opened 2015-01-28
fix picker when close to bottom

MarcGodard:
It now automatically adds bottom padding on pageContent when the picker
is close to the bottom.

Will eventually make a picker keyboard. Maybe.


nolimits4web:
Merged, but also did some modifications as it didn't work always correctly

MarcGodard:
I noticed... I also saw you added it to calendar... i was about to do that.

PR #375, opened 2015-01-28
Tap navbar scroll top

MarcGodard:
Implemented very common tap navbar to scroll to top of page.

Maybe we should add a param to turn this on and off, but many many apps
have it.

### DIFF
--- a/src/js/clicks.js
+++ b/src/js/clicks.js
@@ -2,6 +2,14 @@
 ************   Handle clicks and make them fast (on tap);   ************
 ===============================================================================*/
 app.initClickEvents = function () {
+    function scrollTopHandle() {
+        /*jshint validthis:true */
+        var navClicked = $(this);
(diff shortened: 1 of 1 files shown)


nolimits4web:
Need some improvements:
1) Parametrised and turned off by default
2) Should be also with click on statusbar-overlay
3) If clicked on "center" in navbar, need to check is there a link or not. Because there could be a link or tabs or something similar that shouldn't produce scroll top
4) This should cover all 3 navbar layout types
5) And finally :) Only currently active page should be scrolled top, not hidden ones

MarcGodard:
Awesome, thanks, will make those changes after the weekend.

nolimits4web:
Just found some time to implement it, should cover all usage cases

MarcGodard:
Awesome. Thanks. Been busy with some work lately.

PR #378, opened 2015-02-01
increase the pull distance that triggers pull to refresh

yysun:
Pull to refresh is very sensitive because it is triggered when the page is pull down for 44 pixel. This patch increase the pull distance to 1/4 of the screen.


yysun:
after testing on device, 1/5 seems better. Will submit another pull request.

PR #379, opened 2015-02-01
increase the pull distance that triggers pull to refresh

yysun:
Pull to refresh is very sensitive because it is triggered when the page is pull down for 44 pixel. This patch increases the pull distance to 1/5 of the screen.


nolimits4web:
It is not too sensitive. It is 44px because in most of native iOS apps it is exactly 44px. But i will make it configurable using `data-` attribute

PR #392, opened 2015-02-12
Cannot call method remove/add of undefined

lauborges:
fix "Cannot call method remove/add of undefined" errors


nolimits4web:
How do you get this error? On which element or on which device do you call it?

lauborges:
I get this error just making a require framework7.js in my test file using [jest](https://github.com/facebook/jest)

nolimits4web:
Closing as it was fixed in 1.0.1 release

PR #394, opened 2015-02-14
Responsive Cards

KLVN:
On large screens they have a fixed width so Cards still look like Cards and are not wide stretched.


nolimits4web:
I also thought about it, but this probably should be left on app developer decision, because, for example in most of native iOS apps with cards (Google Plus, YouTube, Twitter) they are still full wide

KLVN:
Ok :)

PR #416, opened 2015-02-28
[virtual-list] fix reinitVirtualList don't work bug

lifei:
use wrong varible cause the code does not work fine


nolimits4web:
Merged, thanks

PR #429, opened 2015-03-11
dom7-ajax: code consistency

f0rmat1k:
Hi! A few code consistency enhancement:
— hasOwnProperty is important because just Object.prototype.someMethod = '...'  can break code execution (It can be just shim or users helpers);
— removed unused variables;
— added var paramsPrefix, it has improved two places at code;

### DIFF
--- a/src/js/dom7-ajax.js
+++ b/src/js/dom7-ajax.js
@@ -3,6 +3,7 @@ var globalAjaxOptions = {};
 $.ajaxSetup = function (options) {
     if (options.type) options.method = options.type;
     for (var option in options) {
+        if (!options.hasOwnProperty(option)) continue;
         globalAjaxOptions[option]  = options[option];
     }
 };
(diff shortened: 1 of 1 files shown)


nolimits4web:
Looks good, did you tested it well, nothing breaks?

f0rmat1k:
No i didn't and don't have possibility at the current time.
Also about 'text/javascript'. It isn't required today.

nolimits4web:
Ok, just wondering can you give me a real world example where not checking hasOwnProperty can break things here in XHR?

f0rmat1k:
Yes, you're right, it's not important today. I've checked es6shim and it's don't write into the prototype.

Only that anti-pattern can bring to unexpected result:
Object.prototype.someCoolMethod = function(){ console.log('!') };
$.serializeObject({one: 'two'}) // => one=two&someCoolMethod=function%20(){%20console.log(%27!%27);%20}

I'l remove checks. But i've seen performance test `for in` vs. `forEach` and forEach was faster and looks better (and no need for 'own' check). I would replace all `for in` to the `Object.keys/forEach/map`. What you think?

PR #435, opened 2015-03-17
Fix #434 - Change VirtualList Callback name

### LINKED ISSUE #434: VirtualList Callback error
In VirtualList callback name is wrong. 

`
if (vl.params.onItemsAfterInsert) vl.params.onFragmentAfterInsert(vl, vl.fragment);
`

gagustavo:
Only change callback name


nolimits4web:
Thanks

Issue #387, opened 2015-02-06
Poor targeting on iPad


johnjamesjacoby:
I have an iPad Air 2 that I use for testing my web-app, and recently the tap targeting has become somewhat erratic and inaccurate. Taps don't register, sidebars don't collapse or swipe away, and it generally feels laggy and unresponsive compared to several weeks ago. The only changes to it in recent weeks have been keeping with F7 updates.

* The iPhone 6+ I use for testing is acting normal.
* I am able to duplicate this laggy behavior using the apps on your example page, such as: http://www.idangero.us/framework7/examples/split-view-panel/
* This happens in both browser and web-app views.
* I do not have a native app to test whether or not this improves anything.
* I am not able to duplicate it using Chrome's developer tools.

nolimits4web:
Not really sure that understand what do you mean. But if you mean that taps sometimes are not working on iPad, you may try to increase `fastClicksDistanceThreshold` app parameter

johnjamesjacoby:
I'll try that. I'll see about recording a demo for you, too.

johnjamesjacoby:
Might it make sense for `var distance` to be:

`app.params.fastClicksDistanceThreshold * app.device.pixelRatio`
 
Since the fastClicks distance is measured in literal pixels, multiplying it by the device pixel ratio will result in a real world translation for the device, rather than white might end up a minuscule number.

Take for example the iPhone 6 Plus with a ratio of `3`. A `20` pixel `fastClicksDistanceThreshold` is still a relatively small margin of touch error.

nolimits4web:
fastClicksDistanceThreshold was set to 10 by default. Anyway, this issue can be fixed by playing with this value. Closing for now

PR #444, opened 2015-03-26
Check if touch moved before removing active state.

johnjamesjacoby:
Fixes issue causing active state on elements to be prematurely removed.

See #387.

### DIFF
--- a/dist/js/framework7.js
+++ b/dist/js/framework7.js
@@ -5922,12 +5922,11 @@
                     trackClick = false;
                     targetElement = null;
                     isMoved = true;
-                }
-                    
-                if (app.params.activeState) {
-                    clearTimeout(activeTimeout);
-                    removeActive();
-                }
+					if (app.params.activeState) {
+						clearTimeout(activeTimeout);
+						removeActive();
+					}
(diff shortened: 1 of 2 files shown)


johnjamesjacoby:
See #387.

nolimits4web:
Cool with first one, but can't accept this one https://github.com/johnjamesjacoby/Framework7/commit/9bd3f5bed49611bf62e4146f96526d0458b58000

It is a bit wrong, during touchmove event we still get distance in rendered pixels, not in physical, so you will just highly increase this distance, effects could be unintentional. So i think, it is better to leave it on developer decision, because you can always pass this parameter manually:
```
new Framework7({
  fastClicksDistanceThreshold: 10 * window.devicePixelRatio
})

johnjamesjacoby:
That makes sense. Whatever you feel is best.

Happy to have that active state stuff sorted!

nolimits4web:
Yep, so will you change the PR or you want me to merge it manually? Btw, do PR only in src/ files, not in dist/

johnjamesjacoby:
Changed. Should be right now.

PR #445, opened 2015-03-27
Make ajax more successful

leecrossley:
Status codes other than 200 are successful. Update to reflect the correct range.

### DIFF
--- a/src/js/dom7-ajax.js
+++ b/src/js/dom7-ajax.js
@@ -26,7 +26,7 @@ $.ajax = function (options) {
         timeout: 0
     };
     var callbacks = ['beforeSend', 'error', 'complete', 'success', 'statusCode'];
-    
+
 
     //For jQuery guys
     if (options.type) options.method = options.type;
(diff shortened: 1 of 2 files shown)


nolimits4web:
I am ok with any 2xx status but not 3xx, they are not really success. They are mostly for browser or direct  request coding on server-side. For example, if you go to `http://www.idangero.us/framework7/kitchen-sink/` and type in console (CORS in browser should be enabled):
```
$$.get('http://idangero.us', function (data){console.log(data)})
```
In network inspector you will see actually 2 requests: 
* one to `http://idangero.us` with 301 status (moved permanently) 
* and then browser will do the second request/redirect to `http://www.idangero.us` and respond with data and 200 status

So in this case we even can't catch the 3xx status. So i think it should be removed from "success" check, as it can produce unwanted results

leecrossley:
With xhr (not network inspector), the 3xx status codes will be automatically dealt with, you will never see a 301 status code for example, it will perform the redirect internally and return the next status code :)

nolimits4web:
Right, so why we need to deal with it in XHR logic?) So can we keep it as `xhr.status >= 200 && xhr.status < 300`? At least, i will feel safe with it

leecrossley:
< 300 is OK also, updated. My main concern was getting failures for 201 etc, which was causing me issues

nolimits4web:
Merged, thanks

PR #446, opened 2015-03-27
Fix undefined selector with 3d panels

leecrossley:
There was an issue when using the 3d panels plugin where selector.nodeType would error when selector was undefined

### DIFF
--- a/src/js/dom7-methods.js
+++ b/src/js/dom7-methods.js
@@ -387,7 +387,7 @@ Dom7.prototype = {
         }
         return this;
     },
-    
+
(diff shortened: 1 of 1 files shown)


leecrossley:
There could be a better fix but I just needed to clear the error for now

nolimits4web:
Thanks, anyway we need to check the selector, but i'll probably will do its check to compare with `undefined`. Btw, how did you get mentioned issue?

leecrossley:
When using 3d panels, it was throwing when the panel opened (latest code f7 code and panels code)

Issue #454, opened 2015-04-03
Ajax form has a invalid Content-Type


yisheng:
The document says:

> Content type will be the same as specified in form's "enctype" attribute. By default (if not specified), it is application/x-www-form-urlencoded

But:

1. If `enctype` attribute is not specified, the `Content-Type` is `null`
2. If `enctype` attribute is specified as `application/x-www-form-urlencoded`, the `Content-Type` is also `null`

![f7-2](https://cloud.githubusercontent.com/assets/352986/6979321/de5a6abe-da14-11e4-8138-9dc14efb7faa.jpeg)

nolimits4web:
We may close it

PR #455, opened 2015-04-03
Fix a bug that causes Ajax Form could not access its "enctype" attribute

yisheng:
Fix Issue #454


nolimits4web:
I will merge it, and add additional check for attribute. But man, you are the first with such issue for a long time, and i can't replicate it in any way. That is something wrong with browser

yisheng:
Thanks for your response! It happened on the lastest Mac Chrome.

PR #484, opened 2015-05-09
Regexp fix for the Calendar widget.

aries1980:
Hi,

There is an issue in the Calendar formatter regexp. The problem is the short name of the months and days can be applied to the long name of as well, which means two different replace is applied to the same string.

My fix expects that after the short name of the day/month there must be a 'non-word' character: whitespace, punctuation, etc.

Let me know if there is a scenario where this pattern does not fit.


nolimits4web:
Merged, thanks!

PR #490, opened 2015-05-19
Fix throttling issue of closeModal() method.

Strate:
If client code tries to close same modal more than once, modal will not shows anymore.
This happens because closeModal creates subscription to `transitionEnd` event twice or more.
I have introduced a flag `f7-modal-closing`, which prevents multiple executions of `closeModal`
method until `closed` event is fired.

Also, this PR moves removing of flag `f7-modal-shown` from `close` event to `closed`, because
modal should be marked as `shown` directly to the end of all animations of closing.

### DIFF
--- a/src/js/modals.js
+++ b/src/js/modals.js
@@ -492,7 +492,7 @@ app.openModal = function (modal) {
         return;
     }
     modal.data('f7-modal-shown', true);
-    modal.on('close', function() {
+    modal.on('closed', function() {
        modal.removeData('f7-modal-shown');
     });
     var isPopover = modal.hasClass('popover');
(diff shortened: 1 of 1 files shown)


nolimits4web:
> If client code tries to close same modal more than once, modal will not shows anymore.

How to replicate this issue? For me, everything works fine

Strate:
@nolimits4web try to call `closeModal()` method more than once, and strictly before `closed` event is fired.

nolimits4web:
In actual dev version, in Kitchen Sink. Open modals page, click on "Popup" button. Then in console type:
```
myApp.closeModal('.demo-popup');
myApp.closeModal('.demo-popup');
myApp.closeModal('.demo-popup');
```
After that you still will be able to open it

nolimits4web:
Or even
```
myApp.closeModal('.demo-popup');
setTimeout(function(){
    myApp.closeModal('.demo-popup');
    myApp.closeModal('.demo-popup');
},150);
```
Still working

PR #493, opened 2015-05-19
Replace "Dummy Link" by "View Source" and add code to show html and j…

mingodad:
This is an attempt to improve the kitchen-sink demo application for newcomers that will show the html and jade files from the current page.

### DIFF
--- a/kitchen-sink/index.html
+++ b/kitchen-sink/index.html
@@ -155,7 +155,7 @@
           </div>
         </div>
         <div class="toolbar">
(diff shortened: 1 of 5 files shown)


nolimits4web:
Sorry, why your feature is only for modals? Then it should be for every page. Also, we don't need to show `jade`, only `html` and `js`

mingodad:
Hello !

It's what I could understand and do so far no other reason than that and as I said it could help other too.

Fell free to make any adjustments if you think it can be worth !

Cheer !

PR #509, opened 2015-06-11
Add `blured` class for navbar/toolbar

FWeinb:
In the iOS9/Safari9 it will be possible to have the blurred look introduced in iOS7 using `-webkit-backdrop-filter`. 
This PR will add basic support for this in the `.navbar` and `.toolbar`. I also modified photoBrowser.js to support a `blurred` option. The kitchen-sink demo is updated accordingly. 

[Here](https://www.youtube.com/watch?v=54MT0fSWO7w) is how it looks on an iPad running iOS9

### DIFF
--- a/kitchen-sink/index.html
+++ b/kitchen-sink/index.html
@@ -148,13 +148,13 @@
     </div>
     <div class="views">
       <div class="view view-main">
-        <div class="navbar">
+        <div class="navbar blured">
           <div class="navbar-inner">
             <div class="center sliding">Framework7</div>
             <div class="right"><a href="#" class="open-panel link icon-only"><i class="icon icon-bars"></i></a></div>
           </div>
         </div>
-        <div class="toolbar">
+        <div class="navbar blured">
           <div class="toolbar-inner"><a href="#" class="link">Dummy Link</a><a href="#" data-popover=".popover-menu" class="open-popover link">Menu</a></div>
         </div>
         <div class="pages navbar-through toolbar-through">
--- a/kitchen-sink/index.html
+++ b/kitchen-sink/index.html
@@ -148,13 +148,13 @@
     </div>
     <div class="views">
       <div class="view view-main">
-        <div class="navbar blured">
+        <div class="navbar blurred">
           <div class="navbar-inner">
             <div class="center sliding">Framework7</div>
             <div class="right"><a href="#" class="open-panel link icon-only"><i class="icon icon-bars"></i></a></div>
           </div>
         </div>
-        <div class="toolbar blured">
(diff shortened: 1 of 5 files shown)


MarcGodard:
Awesome.

nolimits4web:
Yes, finally they delivered something we can use :) I haven't installed yet to check that, but i am interesting about performance. On which iPad you did test it? Because i think to enable it by default for all bars and modals if it works smoothly on iPad 2

FWeinb:
I only tested it on the latest iPad Air 2.

evolutionjay:
Good job！

zb0t:
The Apple website uses this effect as well.

background: rgba(255,255,255,.65),
-webkit-backdrop-filter: blur(20px) saturate(180%)

Works well

chalchis:
someone
it has been tested on a mobile android

ZanderBrown:
Could this be rebased and merged? There seems to be demand for this.

valnub:
I agree, this would be a cool feature. Either this or #1040 should be merged if performance is good.

If it runs smoothly on mid-tier devices, it should be ok since devices are getting faster and faster in the future.

Let's say if it uns smoothly on iPhone 5s (or newer) and iPad Air (or newer), then it should be ok. They both run the Apple A7 chip and will still be supported by Apple in the next version of iOS while older devices might not.

But imho the feature should be **opt-in only**, so devs would have to flip `headerTranslucency:true` or sth like that in config if they're sure that they want to trade in translucency-effect for less performance.

By default, translucency should be turned off in my opinion.

ZanderBrown:
@FWeinb are you aware this PR has merge conflicts?
It would be great if you could resolve those and consider @valnub 's points

J05HI:
This isn't working on Chrome.

ZanderBrown:
@J05HI we know, this is iOS/Safari ONLY.

J05HI:
@ZanderBrown 
I've a solution in mind. 
I'll do a pull request if I have time.

J05HI:
@nolimits4web @ZanderBrown 
Should be closed because it's not a good solution.

ZanderBrown:
Reasoning?

J05HI:
@ZanderBrown
Ios only compatibility

ZanderBrown:
Their is no other way to implement this feature and for the iOS theme only having OSX/iOS comparability isn't a major issue.

And due to the way CSS is applied browsers​ that don't support it will display it as present albeit slightly transparent

valnub:
What if we only add this to iOS stylesheet?

nolimits4web:
Closing as incompatible with current version

PR #536, opened 2015-07-08
fix ellipsis on navbar center text

GuillaumeBiton:
fix for Chrome and IE

### DIFF
--- a/src/less/ios/toolbars.less
+++ b/src/less/ios/toolbars.less
@@ -77,12 +77,12 @@
     left: 0;
     top: 0;
     background: @toolbarBg;
+    .flexbox();
     .hairline(bottom, @toolbarBorderColor);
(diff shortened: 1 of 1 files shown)


nolimits4web:
Yes, it will fix it in Chrome, but there is problem with it because if you put inside anything except it will not behave like it should, and `display:flex` has many advantages on `display:block` especially for something else instead of text, for example buttons

GuillaumeBiton:
OK so I have try and there could be a fix with flex. Need to apply display: flex on .navbar and in .navbar.center remove display lines that enables flex. And it works too

GuillaumeBiton:
Hi, do you want me to close this PR ?

nolimits4web:
I think yes, we can close it for now

PR #545, opened 2015-07-20
speedup picker on android

lihongxun945:
Picker is very wonderful! But it scroll very slow on android devices, even Sumsung galaxy S5.
And I found these code cause the problem in updateItems function:

              col.wrapper.find('.picker-selected, .picker-after-selected, .picker-before-selected').removeClass('picker-selected picker-after-selected picker-before-selected');

              col.items.transition(transition);
              var selectedItem = col.items.eq(activeIndex).addClass('picker-selected').transform('');
              var prevItems = selectedItem.prevAll().addClass('picker-before-selected');
              var nextItems = selectedItem.nextAll().addClass('picker-after-selected');

I removed some of unused code, and it solved the problem(only when rotateEffect is false).

### DIFF
--- a/src/js/picker.js
+++ b/src/js/picker.js
@@ -193,13 +193,12 @@ var Picker = function (params) {
             if (activeIndex >= col.items.length) activeIndex = col.items.length - 1;
             var previousActiveIndex = col.activeIndex;
             col.activeIndex = activeIndex;
-            col.wrapper.find('.picker-selected, .picker-after-selected, .picker-before-selected').removeClass('picker-selected picker-after-selected picker-before-selected');
-
             col.items.transition(transition);
+            col.wrapper.find('.picker-selected').removeClass('picker-selected');
+            if (p.params.rotateEffect) {
+              col.items.transition(transition);
+            }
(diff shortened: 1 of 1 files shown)


lihongxun945:
When there is many items in on col, those code will cause scroll very slow on most android devices.

nolimits4web:
Thanks, i will check it. But these lines:
```js
var prevItems = selectedItem.prevAll().addClass('picker-before-selected');		
var nextItems = selectedItem.nextAll().addClass('picker-after-selected');
```
can't be deleted. You just cropped functionality

lihongxun945:
I find these two classname is not used in css file, and it cause scroll to be very slow on most android devices. And I think you can use other css selectors, such as `.picker-selected ~ .picker-item`, instead of manipulate all items' classname.

lihongxun945:
If rotateEffect is set true,  android device nearly can't scroll the picker. Because you will change all item's transform attr when scroll move.

nolimits4web:
Thanks, i removed these next/prev classes manually. Should be better now

PR #546, opened 2015-07-22
Update dom7-methods.js

dschewchenko:
fix setting/getting element's data- attributes
add more control over shortcut events like in 'on' method

### DIFF
--- a/src/js/dom7-methods.js
+++ b/src/js/dom7-methods.js
@@ -93,10 +93,13 @@ Dom7.prototype = {
             if (this[0]) {
                 // For best perfomance we may to get element's property
                 if (this[0].dom7ElementDataStorage && (key in this[0].dom7ElementDataStorage)) return this[0].dom7ElementDataStorage[key];
-                else if (var dataKey = this[0].getAttribute('data-' + key)) {
-                    return dataKey;
+                var dataKey = this[0].getAttribute('data-' + key);
+                if (dataKey) {
+                    // Set data cache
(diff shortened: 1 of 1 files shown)


adaniello:
:+1:

nolimits4web:
I can agree with extending event shortucts. But not sure about `.data`. First of all i don't think it needs to be mixed with `data-` attributes. There are special methods for that. And data is not always a string we can need to be set anything: array, object, func, .etc. It is not possible to store it in attributes. And i don't really understand this:
```
this[0].dom7ElementDataStorage -> this[0][dom7ElementDataStorage]
```
It will not help to save a lot in compression (just few bites :)), but it such access to properties can cause a big loose in performance as direct access using `.` is much faster

dschewchenko:
I'm fully agree with you. It's my mistake )

dschewchenko:
I have fixed my mistake with data-methods. Now only event shortcuts update in file)

dschewchenko:
And what about deleting attribute in 'removeData' method ?

removeData: function(key) {
        for (var i = 0; i < this.length; i++) {
            var el = this[i];
            if (el.dom7ElementDataStorage && el.dom7ElementDataStorage[key]) {
                el.dom7ElementDataStorage[key] = null;
                delete el.dom7ElementDataStorage[key];
            }
            if (el.getAttribute('data-' + key)) {
                el.removeAttribute('data-' + key);
           }
        }
    },

and I read that 'delete' is not good solution, more good solution is setting variable to null

nolimits4web:
The same, attributes should be intact (adding/deleting) when using `.data` methods

PR #557, opened 2015-07-27
Add canceled event for swipeout elements

DmitryNek:
Event will be triggered after cancellation of swipeout element deletion
confirmation (confirm modal cancellation)

Usage case:

```
$$(el).on('canceled', function () {
      app.swipeoutClose($$(this));
});
```

### DIFF
--- a/src/js/clicks.js
+++ b/src/js/clicks.js
@@ -202,11 +202,15 @@ app.initClickEvents = function () {
                 if (title) {
                     app.confirm(text, title, function () {
                         app.swipeoutDelete(clicked.parents('.swipeout'));
+                    }, function () {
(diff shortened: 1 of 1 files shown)


nolimits4web:
I got your idea by I really don't like the event name. And I guess it is only for closing swipeout element. Maybe it is better to just close it automatically or configure via dat- attribute?

DmitryNek:
Can't get how I can configure swipeout via data- attribute to close it automatically after confirm modal cancellation in current version of fw7. Currently there are no events to catch confirm modal cancellation.

I've found that app.confirm is used only for swipeout in fw7 code, so I've added event to swipeout element.
If you have thoughts on how to name data- attribute, we can do it in this way.

PR #571, opened 2015-08-05
Fix multiple select form storage

log0div0:
The code

```javascript
input.attr('multiple')
```

returns an empty string (which will be interpreted as false) in this case

```html
<select multiple></select>
```


nolimits4web:
Thanks

PR #578, opened 2015-08-11
Fix paramsPrefix for nocache.

pleshevskiy:
when options.cache is false and data not empty then will bug if paramsPrefix will not changed


nolimits4web:
Agree, thanks

Issue #586, opened 2015-08-17
Issue with Pull-to-refresh and on-screen keyboard on iOS, PTR gets stuck


lexaux:
Hi All! 

Thanks for the great framework; it seems to rocket web development for devices, especially fro hybrid apps. 

So the issue for me is as follows: 
1. On iOS 8 (also checked with iOS7, same resutls), we have a pull-to-refresh panel with the input (textbox) on it. 
2. When that input is focused, the on-screen keyboard is shown. 
3. After that, while the keyboard is shown, if I pull the panel down, it just gets stuck (no 'release' really happens), the arrow stays visible. 
   1. However, there's certain threshold to make this happen. 

I have actually wrapped in the simplest possible cordova container one of the demo apps - Hackers news - which features this issue, to see how it behaves. And here is screen recording from it: https://dl.dropboxusercontent.com/u/18312907/framework7_issue_webview_scroll.mov

Would appreciate any help regarding the issue!

Cheers,
Alex.

lexaux:
Done some additional research. 

This is a well-known behavior in the desktop world - basically what happens is that Pull-to-refresh relies on 3 events to operate - touchStart, touchMove and touchEnd. 

In our case, when the keyboard is shown, and user drags an ample distance, the touchEnd event is not received (since touchEnd happens on the keyboard, and not our web view). 

Any ideas how we could fix this? I was thinking of 2 approaches: 

1. Ugly: add a timer, if PTR initiated, collapse it in .5 seconds. Should be really annoying to end users. 
2. Better: somehow detect the webview has lost touch, and either throw in touchEnd or handle it separately. 

Thanks!

ZanderBrown:
What's the status of this?

nolimits4web:
Let's close it for now as there is no more reports about it

PR #588, opened 2015-08-17
Issue #586 proposed fix (terminate pull-to-refresh on soft keyboard).

lexaux:
Terminating pull-to-refresh when there is on-screen keyboard present and touch is moved to the keyboard.

The behavior is still somewhat non-ideal, however no weird artifacts are present. Now it works like this: 
1. Pull-to-refresh initiated on panel 
2. Drag finger to keyboard
3. Safari iOS  8 continues to send `touchmove` events when on keyboard, but no `touchend`! When we detect that `touchmove` coords are lower than `window.innerHeight`, we throw in touchend equavalent event, which terminates ptr exactly as if one did it with their finger.


nolimits4web:
Thanks, but first of all, i can accept commits in `dist/` folder, revert them, as you modify the current release. Also about the issue. I can't replicate it, but i have tried only in home screen webapp and in safari. So is it happens only in Cordova?

lexaux:
hi @nolimits4web, reverted the changes; the pull request is updated accordingly (but it has got 3 commits now). 

For the issue, really I have seen it with Cordova/safari web view on iOS, when the overscroll (bounce effect) is disabled. Not sure about the rest. 

Just checked the http://guillaumebiton.github.io/HackerNews7/ app on an iPhone in different apps, results: 

1. Safari (browser, not web view component). All works well! When you have keyboard on screen, and drag the pull-to-refresh action with the finger finishing touch on top of keyboard, the `touchEnd` event is thrown in, PTR works as expected. 
2. Cordova app (what we use as a container, just pointed it at the Guillaime's app URL: issues. PTR gots stuck if dragging over the keyboard. 
3. There is another app I have on iPhone, called QRReader. Upon reading QR code with URL, it first opens up page in web view, and then there are buttons to open it in safari ios/chrome/etc. So, this app seems to use webview components just as cordova does-  and the issue happens too. 

So I would say, this seems to be an issue of the web view component not throwing the event to JS when touch end over keyboard. Moreover, the fix proposed would only work if the webview **is resized** to the keyboard appearance, which is **not** default behavior on Cordova - I should better add recipe to the issue description in case someone faces the same I guess...

nolimits4web:
Ok, as i guess, the issue probably because of in webview/cordova the keyboard is not a part of webview. I also don't really like this fix as it is a bit hacky and checking window size during touchmove can be a bit slow. Can you try the following, in this place where we add listeners https://github.com/nolimits4web/Framework7/blob/master/src/js/pull-to-refresh.js#L160-L162 try to add one more event handler after:
```
eventsTarget.on('touchcancel`, handleTouchEnd);
```

Maybe (hopefully) it will trigger `touchcancel` event in this case?

lexaux:
Just checked that now, unfortunately no good result. `touchcancel` is not thrown in either, neither are other 'relevant' events like mouseup touchend etc. For some weird reason it continues to throw in 'touchmove' event. 

I've also looked into `setCapture` (outdated) and `requestPointerLock` (desktop version only ,for the newer APIs like dragging out of a window) methods, but no luck.

nolimits4web:
Ok, i will have time check it more deeply on this weekend

ZanderBrown:
Did you manage to take a look?

ZanderBrown:
@lexaux could you resolve the conflicts please?

valnub:
Closed due to no response. The PR is from 2015 as well 🤣

PR #591, opened 2015-08-18
Making the input slider work across browsers (moz & IE styles added)

thomasf1:
We´ve been using the slider styles in a web app and made it work for IE and Firefox.

Please review and merge if you like the changes. Most other styles seem to work in Firefox and IE though :).


nolimits4web:
Thanks, will take a look at it

nolimits4web:
Thanks, i merged it manually as it need to be done for both themes and for colour themes

PR #621, opened 2015-09-14
Ensure disable search bar on input blur

mgesmundo:


nolimits4web:
Sorry, why that? Check for example search in native iOS mail app, it shouldn't be disabled on blur as you won't be able to scroll through search results. As for the case when searchbar overlay is visible (no search) it will be disabled when you click on it. Or did i miss something?

mgesmundo:
Open the search bar page in the kitchen-sink-ios example. While the input search is focused (but empty), the overlay is visible (as expected) and if you press the menu button, it is revealed, but the overlay still visible. If the focus on the search input is lost, I can imagine that the user no longer has interest in search. Obviously the user can tap on the overlay to hide it, but why this additional task? It is only my opinion about the minimum user actions required to complete a task.

nolimits4web:
> If the focus on the search input is lost

Or user just swipes over search results. So it still a question for me

mgesmundo:
If the user swipes over search results, the search bar bar is not disabled by the blur event handler: https://github.com/mgesmundo/Framework7/blob/fix-searchbar/src/js/searchbar.js#L208.

nolimits4web:
But this is still incomplete solution. What if there are links in your searchbar-found links, and if you open another page, and later go back to the page with search results, it will be disabled. Or if you tap close to list-block with results on empty are of the page, it will be disabled again. I guess, we should leave it to app developer decision when he wants to disable it

mgesmundo:
I agree with your consideration: I will add an option to enable/disable this behavior.

ZanderBrown:
Should this be closed? Its 2 years later and it does seem to introduce some odd behaviour

PR #629, opened 2015-09-24
Add new pullstart, pulldown, pullend and pulldone events to Pull To Refresh

mgesmundo:
Your Pull To Refresh component is awesome! I've added some new useful events: I use them to apply a progressive blur an a background when the content is pulled down.


nolimits4web:
Thanks, i have merged them manually and in a bit different moments and with a bit different names :) I choose different moments to differentiate them from basic 'touchstart/move/end` events and to be executed only when PTR container is moved

mgesmundo:
:+1:

PR #636, opened 2015-10-05
dist location update

kovacbb:
error on startup becouse of bad link..


nolimits4web:
Nope, sorry, it is intended as it is used for development purpose and there is a note about it in readme:

But i will change it for future dist releases/tags

PR #639, opened 2015-10-06
Fix statusbar css

mgesmundo:
When a panel (with cover or reveal effect) is with a dark theme, the status bar overlay must be moved with the panel (or the view) to avoid an ugly effect: the panel has the top area with the same color of the status bar.


nolimits4web:
Sorry, but why? It cannot be common solution and may produce more problems/questions. Default/designed behavior supposed to change status-bar color with opened panel, not moving it. Check for example iOS native Facebook app, and the same is done in Kitchen Sink for example

mgesmundo:
Below the answer:
- Fig.1: kitchen sink with default css
- Fig.2: kitchen sink with default css and fixed statusbar color
- Fig.3: kitchen sink with this PR

fig 1 - kitchen sink with default css

fig 2 - kitchen sink with default css and fixed statusbar color

fig 3 - kitchen sink with pr

nolimits4web:
Sorry, i can't agree with this as it is highly depends on app design. What we have now is the most common and easiest to change/tweak way: to change background color or to add transform like in your example. Also there is an idea behind the current realization, if you have half white - half black statusbar background (like on fig 3) then status bar itself can be unreadable as it can be only white or black, so that is why it is easier to handle whole status bar. Actually what is on fig1 is fine, have you checked iOS Facebook app? It and many other native apps behave in the way of Fig1

mgesmundo:
Certainly the look and fell of the app is subjective and of course I full respect your point of view. Every designer/developer can easily customize her css app and it is not a big deal. I've proposed this PR because there is also an idea behind it. Typically when you open a side panel you need a menu item or a short view of some data (list of friends, etc.): this is not a continuous work status and the informations of the statusbar probably are irrelevant at moment. Probably is nice to see a more clear view without the statusbar (using the cordova plugin). In this case the full colored statusbar overlay is ugly.
Facebook is not an idol and its app is not the best app in the world! :-)
Try to see Jawbone UP (https://itunes.apple.com/it/app/up-by-jawbone/id916240764?mt=8&uo=4) and you can see a look and feel like Fig3.

nolimits4web:
I know Jawbone UP, they basically hide status bar when panel opened)). Probably we can make your proposed behavior optional, for example with additional class on status bar. I am just afraid that most of users/devs won't know what is happened with status bar and why it is moving, just trying to avoid such questions in future :D

mgesmundo:
I think that an additional css class is a simple solution :D

ZanderBrown:
@mgesmundo if you could update this PR accordingly that would be great

J05HI:
@mgesmundo
You need to update your merge request (with the additional class)! 
Otherwise it should be closed because of no reaction.

mgesmundo:
Sorry but now I'm too busy :-(

valnub:
@mgesmundo Got time now? :D

nolimits4web:
Closing as incompatible with current version

PR #644, opened 2015-10-08
Fix notifications on Cordova App

mgesmundo:
Sorry for this,
but on a Cordova App, after the first displayed (and closed) notification, the second one throws an error in WebKit (wrong selector).
Thanks.


nolimits4web:
Can't replicate it. What is your cordova/device OS version?

mgesmundo:
Cordova 5.3.3, iPhone 5s with iOS 9.0.2

nolimits4web:
I have cordova 5.1.1 and don't see such issue. Will check with 5.3.3 but actually i don't see any reasons for such errors. Maybe problem on your side, how do you add/close them?

mgesmundo:
I show the first notification on the main page, then I follow a link to another page and I open a new notification. The notification is opened as usual using myApp.addNotification and it is closed using the close button inside the notification itself.

nolimits4web:
Weird, do you use any Cordova plugins? `WKWebView`?

mgesmundo:
Yes: very strange! This is the error:
```
-[WebActionDisablingCALayerDelegate willBeRemoved]: unrecognized selector sent to instance 0x13f539860
```

This is the plugins list:

```xml
    <plugin name="cordova-plugin-whitelist" spec="1.0.0" />
    <plugin name="cordova-plugin-console" spec="~1.0.1" />
    <plugin name="cordova-plugin-splashscreen" spec="~2.1.0" />
    <plugin name="cordova-plugin-battery-status" spec="~1.1.0" />
    <plugin name="cordova-plugin-touchid" spec="~0.3.0" />
    <plugin name="cordova-plugin-camera" spec="~1.2.0" />
    <plugin name="cordova-plugin-ble-central" spec="~1.0.1" />
    <plugin name="cordova-plugin-device-motion" spec="~1.1.1" />
    <plugin name="cordova-plugin-statusbar" spec="~1.0.1" />
```

Tada! The bug in webkit appear when the onClose callback (the same used as onClick callback) contains the app.closeNotification.

nolimits4web:
> Tada! The bug in webkit appear when the onClose callback (the same used as onClick callback) contains the app.closeNotification.

Didn't understand, problem is solved?

mgesmundo:
Removing the doble closeNotification call, I found the real problem: the transitionEnd handler (and also the animationEnd handler) is attached iterating the too large event array:
```js
var events = ['webkitTransitionEnd', 'transitionend', 'oTransitionEnd', 'MSTransitionEnd', 'msTransitionEnd'];
var events = ['webkitAnimationEnd', 'OAnimationEnd', 'MSAnimationEnd', 'animationend']
for (i = 0; i < events.length; i++) {
  dom.on(events[i], fireCallBack);
}
```
and the event handler is called more than one time. Enabling only one of the browser enabled events, there is not problem. I think is better to handle only one of the handler depending of the browser.
Due this consideration, I will close this PR. Do you want I propose a solution for the animationEnd/transitionEnd issue?

nolimits4web:
Did some investigation and 
> the event handler is called more than one time

no, it doesn't called more than once, you can sure in it by putting `console.log` there. So issue can be fixed by using none/single-prefixed event. But i won't do it at the moment as it can break a lot of Androids, as many of them reports falsy support of unprefixed event. 

So let's keep it opened for a while as not really an issue, at least it doesn't break and affect anything except warning message in console

ZanderBrown:
Is it time to close this now?

PR #650, opened 2015-10-11
Fix #649

### LINKED ISSUE #649: Searchbar doesn't show the Cancel button
On the new 1.3.0 release the searchbar doesn't show anymore the cancel button on focus event.

mgesmundo:


nolimits4web:
Thanks for this. I just push a bit different fix to get rid of events at all, and to cover all other cases when it is initialized in hidden panel or popup or wherever else

PR #653, opened 2015-10-12
fix #652

### LINKED ISSUE #652: `showPreloader()` broken in v1.3.0
It can reproduce in kitchen-sink-ios

Tarekajaj:
fixes #652


nolimits4web:
Thanks for noting that. I fixed it with a little bit different way not to output mobal-buttons at all when we don't have any buttons

PR #654, opened 2015-10-13
Scan Dates Range, array accept mixed types (single date + date range)

Tarekajaj:
update for **dateInRange** function that scan date ranges for events, disabled dates etc.. in calendar. now can include date ranges in array to match multiple ranges.
example: 
```javascript
var calendarMultiple = myApp.calendar({
    input: '#calendar',
    events: [
        new Date(2015, 9, 1),
        new Date(2015, 9, 5),
        {
            from: new Date(2015, 9, 10),
            to: new Date(2015, 9, 15)
        },
        {
            from: new Date(2015, 9, 20),
            to: new Date(2015, 9, 31)
        }
    ],
    disabled: [
        new Date(2015, 9, 3),
        new Date(2015, 9, 7),
        {
            from: new Date(2015, 9, 16),
            to: new Date(2015, 9, 19)
        },
        {
            from: new Date(2015, 9, 28),
            to: new Date(2015, 9, 31)
        }
    ],
});
```
will result in
calendarex

I know one could have used a custom function to achieve such result, but this is much more convenient and easy to use and read.

### DIFF
--- a/src/js/calendar.js
+++ b/src/js/calendar.js
@@ -350,9 +350,7 @@ var Calendar = function (params) {
         if (!range) return false;
         if ($.isArray(range)) {
             for (i = 0; i < range.length; i ++) {
-                if (dayDate === new Date(range[i]).getTime()) {
-                    match = true;
-                } else if (range[i].from || range[i].to) {
+                if (range[i].from || range[i].to) {
                     if (range[i].from && range[i].to) {
                         if ((dayDate <= new Date(range[i].to).getTime()) && (dayDate >= new Date(range[i].from).getTime())) {
                             match = true;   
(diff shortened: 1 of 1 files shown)


nolimits4web:
Like the idea and looks good. I am only afraid that this `if (dayDate === new Date(range[i]).getTime())` with passed object can throw an error or warning. Let's place `for-to` check in first and i will merge it, like:
```
if (range[i].from || range[i].to) {
  ...
}
else if (dayDate === new Date(range[i]).getTime()) {
  ...
}

Tarekajaj:
Done

nolimits4web:
Thanks, merged

PR #663, opened 2015-10-15
Add support for DOM content in router.preprocess

caiiiycuk:
I wish to use Angular + Framework7 together. For that i should compile pages content with angular.$compile. I use router.preprocess to achive this. But $compile returns DOM element with angular scope, so i can't return HTML text from preprocess function. In this PR router is modifed with additional check. If content is not string then router appends content and don't use innerHTML field.

Example of my preprocess function:
```coffeescript
    @framework7 = new Framework7
      swipePanel: 'left'
      preprocess: (content, url, next) =>
        compiled = @compile content
        compiledContent = compiled @scope
        compiledContent
```

### DIFF
--- a/src/js/router.js
+++ b/src/js/router.js
@@ -262,7 +262,7 @@ app.router._load = function (view, options) {
 
     // Parse DOM
     if (!pageName) {
-        if (url || (typeof content === 'string')) {
+        if ((typeof content === 'string') || (url && (typeof content === 'string'))) {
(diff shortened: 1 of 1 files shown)


caiiiycuk:
I don't know why but this PR breaks back navigation...

caiiiycuk:
Sorry my mistake, all works fine...

nolimits4web:
Ok, seems to be everything is good. Merged, thanks

PR #665, opened 2015-10-16
Router fixes

robert-blankenship:
If `app.params.pushState` and `app.params.pushStateSeparator` were enabled, an additional hashtag would be added to the url of the page, e.g., `data-page='main' ---> ##main`, when it should be `#main`.

When moving back in the history, the contentCache would sometimes be cleaned even though the page was still present farther back in the history.


robert-blankenship:
@nolimits4web I've also added a check for `app.params.preventPushStateOnLoad` in the view onLoad.

Let me know if you need anything.

nolimits4web:
Hi, looks interesting, but for some reason i can't replicate it to see such issue. Maybe you can give me small guide on how can i replicate it?

robert-blankenship:
Sure; I'll set something up and link it to you here. I should be able to let you know by Tuesday.

nolimits4web:
Ok, thanks

robert-blankenship:
test for dist:
http://robertblankenship.xyz/f7/?throwFailures=true&useRouterFixesBranch=false

test for PR branch:
http://robertblankenship.xyz/f7/?throwFailures=true&useRouterFixesBranch=true

The tests are written with Jasmine.

spec file:
http://robertblankenship.xyz/f7/spec.js (compiled JavaScript) 
http://robertblankenship.xyz/f7/spec.coffee (CoffeeScript)

nolimits4web:
Ok, i see what you mean. As for doubled hashes, the additional hash in front of page name was added intentionally. 

For example, with your fix. If i open an app (first time, when there is no any states in history) by url `myapp.com/!#/about`. How will i know, should i load inline page or make Ajax request? Need to think here how to cover such cases. Optionally i can just add parameters to specify full URL rules for such pages, so by default it can be like `#{{name}}`.

As for `preventPushStateOnLoad`. This parameter is intended for different case. For example in mobile Safari, it fires falsy pop state event for some reason. And setting this parameter to `true` blocks `popstate` event handling until window is loaded to fix this issue. But i understand you, its name is a bit confusing, probably it can do what you purposed and rename the current parameter to different

robert-blankenship:
If a page is opened using AJAX, then the url would be `myapp.com/#/about.html`. So there shouldn't really be any conflict.

Here is an idea to handle the onload state:
1) Check `history.state`
2) Check for inline pages
3) Check for pages templates
4) Do AJAX as a last resort.

The benefit here is that 1-3 can be done quickly and synchronously.

https://github.com/robert-blankenship/Framework7/commit/35da0584f05d54acff749fc97c5ed7f7a9f5f9f0
I don't think we disagree about anything in this commit, should I make a separate PR for it?

nolimits4web:
I am ok with https://github.com/robert-blankenship/Framework7/commit/35da0584f05d54acff749fc97c5ed7f7a9f5f9f0 . Yes, you can make separate PR for it, or i can merge it manually, just let me know

robert-blankenship:
Sorry for the delay, I see that you've already added the commit. Thanks! I'll close this PR.

PR #670, opened 2015-10-22
Fix dom7 parent

usagizmo:
If the case of `this[0].parentNode => null`, to become `parents => [null]`.
This Error has occurred when using an Dynamic Pages.

From

```
// Accordion
if (clicked.hasClass('accordion-item-toggle') || (clicked.hasClass('item-link') && clicked.parent().hasClass('accordion-item'))) {
```


nolimits4web:
Merged, thanks

PR #672, opened 2015-10-23
!a: showFilteredListOnly parameter to virtual-list

kozmanbalint:
Set this parameter to true and virtuallist will show items only if there's a filter, but will not show the items of the unfiltered list.
I started a forum thread on this topic (http://www.idangero.us/framework7/forum/#!/getting-started:how-to-hide-unfiltered-item), but as there were now replies for some time, I thought it was the first time this idea came up, so I decided to implement it myself. Please, review if you find it useful and correct.

### DIFF
--- a/src/js/virtual-list.js
+++ b/src/js/virtual-list.js
@@ -6,7 +6,8 @@ var VirtualList = function (listBlock, params) {
         cols: 1,
         height: app.params.material ? 48 : 44,
         cache: true,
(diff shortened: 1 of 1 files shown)


nolimits4web:
Hi, thanks. But this fix was a bit wrong. I've renamed parameter and added correct fix for that

kozmanbalint:
Thanks!

PR #673, opened 2015-10-23
Fix condition for bars reappearance

jie-nu:
This fix prevents bars to be reappeared even if showBarsOnPageScrollEnd is false when the screen scrolls to end and bounces.


nolimits4web:
All good, thanks

PR #711, opened 2015-11-13
Ignore cache on .xhr fix

sznowicki:
IgnoreCache flag should not cause the clear cache functionality to be ignored. Otherwise UI gets the old cached version on next request.


nolimits4web:
Merged, thanks

Issue #714, opened 2015-11-17
Improvement - Badge colors


lleevvyy:
Hi, I propose to add ability to colorize a badge. Very simple change in colors css is required.

nolimits4web:
Merged, thanks

PR #715, opened 2015-11-17
Badge colors, #714

lleevvyy:


nolimits4web:
Thanks, i think we can add this. But you can actually already specify badge color by specifying its `bg-` class like `<span class="badge bg-green">1</span>` :)

lleevvyy:
Thanks :-)

PR #772, opened 2015-12-30
add onlyImage field to messages properties

paulhovey:
I came across this problem when adding emojis to a message field - emojione converts into images, which are embedded in the message text.  However, the css for message-pic doesn't play well with text, which makes sense.  I thought it appropriate to add in another field to determine not only if there is an image anywhere in the text, but if it is the beginning (and assumed to be the only significant text, so we can apply message-pic).  For me, I disable autoLayout and use onlyImage to add message-pic class to the messages display like this, and it works well.

### DIFF
--- a/src/js/messages.js
+++ b/src/js/messages.js
@@ -102,6 +102,7 @@ var Messages = function (container, params) {
             props.type = props.type || 'sent';
             if (!props.text) continue;
             props.hasImage = props.text.indexOf('<img') >= 0;
+            props.onlyImage = props.text.indexOf('<img') == 0;
             if (animate) props.position = method === 'append' ? 'bottom' : 'top';
 
             newMessagesHTML += m.template(props);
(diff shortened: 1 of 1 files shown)


nolimits4web:
Hi, thanks for PR but i didn't really get the issue. Can you provide some example or how to replicate it?

paulhovey:
Sure, it was to fix an issue I was having and wasn't sure it warranted a generic change but wanted to leave that up to you.  In my original comment I mentioned having an emoji picture inline with text caused the message text field to be displayed incorrectly because it found an image within the text field.  Really, it seems that message-pic should be applied only if there is a picture, and not if there is text other than a picture.  I'll try to get screenshots later.

paulhovey:
hasimage ios
onlyimage ios

I hope the attachments make sense.  The first one uses hasImage, and the background turns to white and you can't see the text.  The second image uses onlyImage, and the background is present and the text visible.

paulhovey:
I made another commit to show where the issue is affected, in the messages template.  Sorry for not being clearer before.

PR #773, opened 2015-12-30
Add index for contacts list

lihongxun945:
Add an alphabetical index scrollbar for contacts list. I am not good at designing, so the style may not look very nice.


nolimits4web:
This looks interesting. But why, if there is a https://github.com/boynet/Framework7-Indexed-List-plugin plugin already exist. I just didn't want to move it to F7's core because it is pretty rarely used. Maybe there is a more sense to PR it to https://github.com/boynet/Framework7-Indexed-List-plugin plugin?

lihongxun945:
OOOOOH~~, I really do not know that there is another plugin exist. I write this plugin in my own project by myself.

PR #828, opened 2016-02-03
Push state on load

Dr1ks:
App parameters to disable push state on load. By default it true, so standart behavior would not changed.


nolimits4web:
Merged, thanks

PR #841, opened 2016-02-11
needed to allow pickers to be displayed in popups using javascript by…

lbaker2:
… exposing openOnInput. Also prevented event bubbling, so other elements may open the picker on click without triggering HTML click

I noticed there was a problem calling app.picker.open() if my picker's input is inside of a popup, but worked fine when clicking the input itself (all that is missing is the overlay, but that is not a problem). I added the openOnInput method to the picker so that the animation and correct positioning can happen through javascript. I also added e.stopPropagation, so that another element, such as an icon, can receive a click event and delegate handling the event to the picker without triggering HTML click (which closes the picker).

This is a workaround, but a question I had while writing this was why onInputOpen and picker.open were two separate methods?

Thanks,
Lawrence

### DIFF
--- a/src/js/picker.js
+++ b/src/js/picker.js
@@ -437,11 +437,14 @@ var Picker = function (params) {
 
     // Input Events
     function openOnInput(e) {
-        e.preventDefault();
+        if(e){
+            e.preventDefault();
+            e.stopPropagation();
+        }
         if (p.opened) return;
         p.open();
         if (p.params.scrollToInput && !isPopover()) {
-            var pageContent = p.input.parents('.page-content');
+            var pageContent = p.input.parents('.page-content, .popup');
             if (pageContent.length === 0) return;
 
             var paddingTop = parseInt(pageContent.css('padding-top'), 10),
@@ -507,6 +510,8 @@ var Picker = function (params) {
     }
 
     p.opened = false;
+    p.openOnInput = openOnInput;
+    p.closeOnHTMLClick = closeOnHTMLClick;
     p.open = function () {
         var toPopover = isPopover();
 


nolimits4web:
Hmm, interesting. Can you guide me a bit how to replicate the issue?

fox000:
I believe that my problem is similar.

This code doesn't work:
``` javascript
$$('#my-button').on('click', function () {
    
    var pickerDevice = myApp.picker({
        input: '#selected-device',
        cols: [
            {
                textAlign: 'center',
                values: ['iPhone 4', 'iPhone 4S', 'iPhone 5']
            }
        ]
    });
    pickerDevice.open();

});
```

Only clicking on input tag can open picker. What is interesting is that, clicking #my-button add picker to DOM, but picker is not visible. And more interesting is that on big screen (iPad) it opens it as popover, but as i mentioned earlier, in iPhone it is invisible.

ZanderBrown:
@lbaker2 could you please address @nolimits4web 's request and clear up the merge conflicts please?

lbaker2:
It's been a long time since I've looked at any of this, but I'll do my best to explain what was happening. I don't know how/where to view the merge conflicts and my suggestion is not to accept this pull request, but consider it as acknowledgement of an underlying issue and reason to refactor this part of the Framework7 Picker code (mainly point 1 below).

There were two objectives 

**1)** I wanted to open a picker by clicking a label or icon.

My initial attempt looked similar to @fox000 's example, as the docs state picker.open() opens the picker. This revealed that picker.open() did not handle adding padding to the parent container and scrolling to the input. Again, @fox000 mentioned this behavior. This is why I exposed openOnInput on the app.picker component. By calling picker.openOnInput directly I got all the styling and was happy.

Example
```
var picker = app.picker({
      input: '#picker-device'
      //...
    });

//...

// - hacky, but shows our picker when superduper icon is clicked
$$('.superduper-icon').on('click', function(){ picker.openOnInput(); }

// - picker is off screen :(
$$('.superduper-icon').on('click', function(){ picker.open(); } 

// - alternative without my changes for fox000 to try
// - relies on knowledge of undocumented property 'picker.input' which is a Dom7 object.
// - picker.input is derived from $(picker.params.input)
$$('.superduper-icon').on('click', function(){ picker.input.trigger('click'); }
```


Taking a glance at what is on the master branch currently, I believe the original behavior I described is still happening. openOnInput is only called via a click event handler bound the picker's input (picker.js:494).

**Lingering Question**

In my original comment I asked why onOpenInput and picker.open are two separate methods. This is really asking, when would we want to open (render) the picker without adding any of the styling to make it display properly?

**2)** I wanted to show the picker from within a popup

Clicking a picker's input from within a popup could not find the popup to calculate the correct padding to add. This was fixed by add an additional selector to openOnInput when determining new styles to display the picker within the user's view.

ZanderBrown:
@nolimits4web any thoughts?

cyberbeat:
I experienced the same issues with the datepicker on a hidden input field.

nolimits4web:
It is fixed in v2

cyberbeat:
I know it does not belong here, but: great work, what you have done for v2! It seems you have fixed all the issues I had with v1 (app layout, components,..). I hope to see some v2-docs soon!

PR #898, opened 2016-03-22
fixes for nested views: issue #1082 and issue on virtual-list.js

reecube:
- Now it checks if the vl.listBlock element is visible before it does the resizing- and rendering-call. Like this, the virtual list won't be overwritten with wrong content when it's not visible. This happens when you have a routing with more than 1 level because just the current back-page keeps visible.
- Also fixed the following issue: https://github.com/nolimits4web/Framework7/issues/1082

### DIFF
--- a/src/js/router.js
+++ b/src/js/router.js
@@ -300,7 +300,7 @@ app.router._load = function (view, options) {
         oldPage = pagesInView.eq(pagesInView.length - 1);
     }
     else {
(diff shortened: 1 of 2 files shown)


ZanderBrown:
what happened there

reecube:
@ZanderBrown what happened where? If you mean the closed and reopened, that was an accident.

PR #925, opened 2016-04-06
Fix an issue in router._back function

### LINKED ISSUE #861: allowDuplicateUrls option don't work when going back
I have a Template7 Ajax page that is loaded sequentially multiple times, with diferent context each time.
I'm using **allowDuplicateUrls** option to allow loading the same url time after time.
This works perfectly loading new pages, but fails going back.

I could make it work changing line 937 in router.js.
before:
`if (url && url === view.url || pageName && view.activePage && view.activePage.name === pageName) {`
after
`if (!view.params.allowDuplicateUrls && (url && url === view.url || pageName && view.activePage && view.activePage.name === pageName)) {`

This works only when setting back option **force: true**.
I'm not sure if what I did is right or if it's enough, 
I'm new to Framework7 so I suppose I'm missing something.

Thankls for your help.
Leo

LeoRodz:
Fixes #861.
Add conditional `&& !view.params.allowDuplicateUrls` when force==true and the url or pagename is the same.
In this way, the condition follow the same criteria used in router.load function, line 592.

### DIFF
--- a/src/js/router.js
+++ b/src/js/router.js
@@ -934,7 +934,7 @@ app.router._back = function (view, options) {
         }
     }
     else {
-        if (url && url === view.url || pageName && view.activePage && view.activePage.name === pageName) {
+        if ((url && url === view.url || pageName && view.activePage && view.activePage.name === pageName) && !view.params.allowDuplicateUrls) {
             view.allowPageChange = true;
             return;
         }


ZanderBrown:
@LeoRodz could you please fix the conflicts and explain why / what you have done here?

LeoRodz:
The modification fixes the issue #861 (there is a demo app showing the problem).
The modificatoin has sense because it follows the same logic implemented in app.router.load function at line 667.

ZanderBrown:
Okay makes sense but could you update the PR as it currently has conflicts

valnub:
@LeoRodz Status?

nolimits4web:
Closing as incompatible with current version and probably not needed anymore

PR #930, opened 2016-04-10
Catch formFromJson event to resize textareas on load

steadweb:
Added `.on('formFromJSON')` listener to automatically resize `resizeable` textareas on load.


ZanderBrown:
This looks like it might be good

ZanderBrown:
Thanks @valnub hadn't notice the conflict. @steadweb could you resolve the conflicts please?

nolimits4web:
It is not needed as form:fromdata already triggers `change` event on inputs

PR #964, opened 2016-05-05
build -> dist fix

ionas:
This fixed the demo pages for me.


nolimits4web:
There is info about it in readme:

ionas:
My suggestion then is (as an easy hack) to have CSS files linked in via build and dist
Should I set that up as a PR?

PR #965, opened 2016-05-05
build -> dist fix

ionas:
same for android


nolimits4web:
There is info about it in readme:

PR #1003, opened 2016-05-31
Fixed incorrect vrirtual list height

Edrem:
This is a small bugfix, sorry I did not open an issue first but the solution seemed trivial and only seems to be an oversight.

When using multiple columns virtual list does not properly change it's height. This change fixes it to always round up to the current column height and prevent weird layouts when a row is not entirely filled with columns.


nolimits4web:
Sorry for late review. Merged, thanks

PR #1021, opened 2016-06-12
Added trigger hide to tabs

EmilioBravo:
Added a trigger('hide') to oldTab so It would be possible to hook to the tab hide event in the same way you can hook to the show event now.


nolimits4web:
Cool, thanks!

Issue #1080, opened 2016-07-29
Statusbar cutting in iOS simulator after build


frisone:
This is a (multiple allowed):
- [x] bug
- [ ] enhancement
- [ ] feature-discussion (RFC)
- Framework7 Version: nobutakaoshiro:framework7-ios-material.
- Platform and Target: IOS.
### What you did

Resolve when statusbar show, the toolbar displayed cutting
### Expected Behavior

Statusbar don't cutting when statusbar show.
### Actual Behavior

When statusbar show, the toolbar displayed cutting.

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/framework7)
for that or join the #framework7 channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

valnub:
Hi, sorry for the late answer. I didn't really understand your problem. Could you provide a screenshot? Thanks.

scriptPilot:
With [App Framework demo](http://app-framework.scriptpilot.de/) i have the same issue on iPhone SE.

![image 7](https://cloud.githubusercontent.com/assets/19615586/22228085/1315c37a-e1cf-11e6-9090-307f2fc9954a.png)

scriptPilot:
PS: Only after pinned it to the homescreen

valnub:
Hm, yes I can confirm the problem.

Maybe this helps? https://framework7.io/docs/status-bar.html

Did you try this?
```
<div class="statusbar-overlay"></div>
```
and

```
.statusbar-overlay {
    background: transparent;
}
```

Also, you might wanna try this: https://muut.com/framework7#!/framework7/bugs-and-issues:status-barlogin-screen-iss

`<meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">`

scriptPilot:
@valnub - Thank you, fixed it for F7, but not for F7 - I opened a new ticket nolimits4web/Framework7-Vue#71

PR #1081, opened 2016-07-29
resolve issue #1080

frisone:

Always follow the [contribution guidelines](https://github.com/nolimits4web/Framework7/blob/master/.github/CONTRIBUTING.md) when submitting a pull request.

### DIFF
--- a/src/less/ios/statusbar.less
+++ b/src/less/ios/statusbar.less
@@ -8,6 +8,9 @@ html.with-statusbar-overlay body{
     .panel {
         padding-top: 20px;
     }
+    .toolbar{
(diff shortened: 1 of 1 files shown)


ZanderBrown:
I'm guessing from the title this is related to #1081 but it would be good if you explained this in the message instead of leaving the default one saying you should explain

ZanderBrown:
Could you please expand you point?
Consider pointing code snippets here, on [Codepen](https://codepen.io) or on [JSFiddle](https://jsfiddle.net).
Make sure you have read the [documentation](http://framework7.io/docs/) correctly and followed any instructions it may give.

ZanderBrown:
#1080 was resolved as incorrect usage, thus this is superfluous

PR #1138, opened 2016-09-26
Router bugfix - Default target for back button

### LINKED ISSUE #1131: 'back' link doesn't load fallback page from 'href'
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Framework7 Version: 2b0fa3a5b171b1a3ca0ed2f3d6f97d6a6cd108ac.
* Platform and Target: iOS, ANDROID.

### What you did
Adding the 'back' class to a link with a 'href' attribute on a first page doesn't open the fallback 'href' url .

### Expected Behavior
As the [doc](http://framework7.io/docs/pages-ajax.html#going-back-in-navigation) says:
> If there are pages in navigation history Framework7 will always ignore url in "href" attribute
> If there are no pages in navigation history (for example, when you have "back" link on first home page) Framework7 will load page from "href" attribute

```js
// https://github.com/nolimits4web/Framework7/blob/master/src/js/router.js#L1008
url = options.url = view.history[view.history.length - 2];
// Should not override options.url and do:
url = view.history[view.history.length - 2] || options.url;
```

### Actual Behavior
It doesn't go back to the fallback page.

valnub:
Fixes #1131 

Don't know if this has any other side effects though. Currently the fix seems to work fine for me.

### DIFF
--- a/src/js/router.js
+++ b/src/js/router.js
@@ -1005,7 +1005,7 @@ app.router.back = function (view, options) {
         return;
     }
     if (!force) {
-        url = options.url = view.history[view.history.length - 2];
+        url = view.history[view.history.length - 2] || options.url;
         if (!url) {
             view.allowPageChange = true;
             return;


nolimits4web:
Looks good, thanks!

PR #1165, opened 2016-10-12
iOS10 gets device.iphone failure

wapznw:
iOS10 browser Useragent iPhone OS replaced iOS.


nolimits4web:
Merged manually! Thanks!

PR #1170, opened 2016-10-18
Fix material chips direction in RTL mode

Siton-Systems:
Before in RTL the design look broken.

See codepen:
http://codepen.io/siton/pen/YGJVrK?editors=1100


nolimits4web:
Merged, thanks

Issue #1166, opened 2016-10-12
Ajax error callback is always called  (after success callback) if I set timeout option


IvanPaiano:
This is a (multiple allowed):
- [x] bug
- Framework7 Version: V1.4.2.
- Platform and Target: All
### What you did

$$.ajax({
    url: 'operation/update',
    data: {id: 10},
    cache: false,
    method: 'POST',
    dataType: 'json',
    async: false,
    timeout: 10000, //10 sec
    success: function (data, status, xhr) {
        alert("OK!");
    },
    error: function (xhr, status) {
        alert("KO!");
    }
});

I get "OK!" alert and after 10 seconds I get "KO!" alert...
url returns a correct JSON .
Perhaps it is not properly reset timeout timer...

Thank you.

macdonst:
@IvanPaiano I think I have a fix for this issue. More detailed description in the PR.

PR #1179, opened 2016-10-26
Issue #1166: Ajax error callback is always called (after success callback) if I set timeout option

macdonst:
This is a bug fix for issue #1166. User @IvanPaiano was send a synchronous POST with a timeout. They way the code was written is the send would happen first, block for a result, fire the success callback and then the timeout would be created. 10 seconds later the timeout would fire and the error callback would be invoked.

This wouldn't happen with an async call (or at least would be less likely) as the xhr.send() call is not blocking.

The fix is pretty simple, move xhr.send() to after the setup of the timeout.

### DIFF
--- a/src/js/dom7-ajax.js
+++ b/src/js/dom7-ajax.js
@@ -241,10 +241,6 @@ $.ajax = function (options) {
     fireAjaxCallback('ajaxStart', {xhr: xhr}, 'start', xhr);
     fireAjaxCallback(undefined, undefined, 'beforeSend', xhr);
 
-
-    // Send XHR
-    xhr.send(postData);
-
     // Timeout
     if (options.timeout > 0) {
         xhr.onabort = function () {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks!

macdonst:
@nolimits4web no problem, hoping to help out more. Especially on the F7 + Vue project. Can you add more issues so I can get a sense of how to help out or is there a mailing list I can join?

nolimits4web:
@macdonst will accept any help with pleasure especially for F7+Vue, the problem is that it is new-comer, so no new issues until people and us start to use it. I'm personally just started to use it first time on my client's project, so will definitely will find some bugs. But it is just has to be used to find more bugs and uncovered use cases

macdonst:
@nolimits4web okay, cool. I should try to convert my demo app over to it to see where things fall down.

nolimits4web:
@macdonst the only question for me at the moment which i don't really understand how to combine with F7 api are overlays. If you have any thoughts https://github.com/nolimits4web/Framework7-Vue/issues/4

PR #1180, opened 2016-10-26
Ios filled buttons vertical-align problem fixed.

ibrahimyilmaz7:
Ios buttons have 2px lover line-height from height. I think it's because of borders (1px top, 1px bottom). When we set border:none; to filled buttons its cause vertical align problem. 

Using border-color: transparent instead of border: none; solves problem.


nolimits4web:
Merged, thanks!

PR #1214, opened 2016-11-21
Fixed issue with search bar and grouping in smart selects.

trancee:
When using smart select with optgroup and the search bar, whenever you start to type a letter in the search bar, the following error will occur:

```
smart-select.js:391 
Uncaught TypeError: Cannot read property 'toLowerCase' of undefined(…)
searchByItem @ smart-select.js:391
Searchbar.s.search @ searchbar.js:211
(anonymous function) @ searchbar.js:164
```

I have fixed this issue by checking if there is a text value in the currently iterating item.


nolimits4web:
Merged, thanks!

PR #1217, opened 2016-11-24
Fixed double click on element with ripple effect

ximiksergey:


nolimits4web:
Thanks! Merged manually as dist files shouldn't be changed

PR #1224, opened 2016-12-01
Framework7 Ajax Method: Posting json data which contains "&" sign.

ibrahimyilmaz7:
Why are we replacing "&" with new line character on ajax json data?

_data.replace(/&/g, '\r\n');
This replace operation causes json error on the server. Why we need this replace operation?

## With current replace method:
```
var _data = "Tarih & Saat";
_data.replace(/&/g, '\r\n');
```

**Result**
> "Tarih
> Saat"

in JSON at position 927token
    at Object.parse (native)
    at parse (C:\Users\ibrahimyilmaz\workspace\appcreator\node_modules\body-parser\lib\types\json.js:88:17)


nolimits4web:
Thanks, fixed!

PR #1230, opened 2016-12-05
Added subtitle for navbar center

etiennea:
Added subtitle for navbar center in iOS
Like in whatsup for example

Picture

### DIFF
--- a/src/less/ios/toolbars.less
+++ b/src/less/ios/toolbars.less
@@ -105,6 +105,15 @@
         .flex-shrink(10);
         .flexbox();
         .align-items(center);
+        .subtitle {
+	        color: gray;
+	        display: block; 
+	        font-size: 10px;
+	        position: absolute;
+	        bottom: -14px;
+	        text-align: center;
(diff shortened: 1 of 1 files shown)


ZanderBrown:
Can you provide an example of this being used in an existing app as F7 should keep at close the the design guidelines as possible

etiennea:
Whatsup and the official messages app too.

Although in the  official messages app the navbar become a few pixel higher.

ZanderBrown:
@etiennea good call

etiennea:
Twitter also uses it a lot. Many many apps use this pattern.

JosephShenton:
image
Here is a preview of it i made some changes to it to look better also converted it to css but here is the css surely you know how to put it back into less

```
.center .subtitle {
    color: white;
    display: block;
    font-size: 12px;
    position: absolute;
    bottom: -13px;
    text-align: center;
    width: 100%;
    font-weight: bold;
}
```

nolimits4web:
Merged, thanks!

PR #1249, opened 2016-12-19
Fixed issue with empty Material Tabbar.

trancee:
If we use the Material Tabbar and set it to scrollable, but the Tabbar itself is empty, an error will occur because the active tab bar does not exist. This fix should prevent Framework7 from throwing an exception and continue to work as expected.


nolimits4web:
Merged, thanks!

PR #1252, opened 2016-12-22
Update virtual-list support DOM7

jinzhe:


nolimits4web:
What do you mean by "support Dom7"? It is already supported. If in your example, you will pas HTMLElement as listBlock parameter nothing will work at all

ZanderBrown:
Along with a description this pull request is broken according to travis. Please fix this.

src/js/virtual-list.js
```diff
@@ -18,7 +18,7 @@ var VirtualList = function (listBlock, params) {
 
     // Preparation
     var vl = this;
-    vl.listBlock = $(listBlock);
+    vl.listBlock = typeof listBlock=="string"?$(listBlock):listBlock;
     vl.params = params;
```
ZanderBrown: What is this doing and why?

You mentioned DOM7 but this isn't part of the DOM7 library

ZanderBrown:
Unexplained, has conflicts and no response has been given.

PR #1261, opened 2016-12-28
Improve lazyloading images

kuldipem:
**New Feature** for **Image LazyLoad**

On lazy loading a image, it is good idea to show placeholder/default image if somehow `data-src` or `data-background`  src image couldn't get loaded.


nolimits4web:
Merged, thanks

PR #1270, opened 2017-01-05
make parseQueryUrl more compatibility

hwygy001:
when i update the framework7 to the version 1.5.2, i found the parseQueryUrl has been changed. I learn this change for fix the hash status.('http://www.baidu.com?name=li&age=18#tab1').
**But for a normal occasion,** when i set the pushState=true in app init, all my url will become this style (http://127.0.0.1:8080/#!/views/test-query.html?abc=333)， the parseQueryUrl can't get any params. so i do this request, and this change will fix more occasion.

1. can parse the hash status, such as **http://www.ttt.com?name=li&age=18#tab1**
2. fix the push status true status, such as **http://127.0.0.1:8080/#!/views/test-query.html?abc=333**
3.when parseQueryUrl without params ,it will get the location.href as default.

### DIFF
--- a/src/js/dom7-utils.js
+++ b/src/js/dom7-utils.js
@@ -1,4 +1,8 @@
 // DOM Library Utilites
+/**
+ * searchStr: the url need to be parse
+ * name: the key in parse object , defalut empty
+ */
 $.parseUrlQuery= function (searchStr, name) {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Merged, thanks

Issue #897, opened 2016-03-22
context lost on back when level > 1


reecube:
I'm using the current release 1.4.2 of framework7.

I have an application with a master-detail view where you could go deeper than level 1. Works fine until you click on the back button. When you try to go back framework7 looses the context and is not able to render the page correctly.

For example (all pages are in #view-main):
# page-master (context: movies)
# page-detail-movie (context: movie.@index)
# page-detail-actor (context: actor.@index)

If you are on the master page and select a movie, the app navigates to the movie-detail page with the given context and then you can select an actor to navigate to the actor-detail page. Everything is working fine here. But then you click on "back" from actor-detail page and the app navigates to movie-detail page but the context is lost, this means there is no data to show.

PS. I'm using Ajax-Pages and Template7 to render them with the given context.

nolimits4web:
It depends on how do you pass context there. I can recommend you to enable `domCache` for this View

shuizhongyueming:
@nolimits4web 

I have same problem for this. And I was using the release 1.4.2 also.

In my project, there is a `gameDetail` page that is load dynamically

```javascript
        mainView.router.load({
            url: $$curr.data('href')+'?id='+gameId,
            context: Template7.data[api.DATA_TYPE_GAME][gameId]
        });
```

and in the `gameDetail` page, you can load another `gameDetail` page with different gameId.

so, the history will be some like this:

```javascript
[
  'game-detail.html?id=1',
  'game-detail.html?id=23',
  'game-detail.html?id=14',
  'game-detail.html?id=56',
  'game-detail.html?id=6',
  ...
]
```

now, if the page is active with `game-detail.html?id=6`. then, press the back button on the nav.

the page will trans to the `game-detail.html?id=56`, and because the page `game-detail.html?id=56` is cache in the dom, so it will show normally.

But because the history's length is bigger than 1, so the framework will try to preload page 'game-detail.html?id=14'

```javascript
// router.js afterBack

// Preload previous page
if (view.params.preloadPreviousPage) {
    if (view.params.domCache && view.history.length > 1) {
        var preloadUrl = view.history[view.history.length - 2];
        var previousPage;
        var previousNavbar;
        if (preloadUrl && view.pagesCache[preloadUrl]) {
            // Load by page name
            previousPage = $(view.container).find('.page[data-page="' + view.pagesCache[preloadUrl] + '"]');
            if (previousPage.next('.page')[0] !== newPage[0]) previousPage.insertBefore(newPage);
[...]

PR #1278, opened 2017-01-11
Fixed issue with missing context when router went back.

trancee:
This fixes issue #897 when there are multiple pages in the history and the router went back one page, it will then preload the second last page in the history but without setting the context. I have used the contextCache to restore the context of that page. Please review and merge if it looks good. It works for me.

### DIFF
--- a/src/js/router.js
+++ b/src/js/router.js
@@ -1088,6 +1088,7 @@ app.router.back = function (view, options) {
     function proceed(content) {
         app.router.preprocess(view, content, url, function (content) {
             options.content = content;
+            options.context = view.contextCache[options.url];
             app.router._back(view, options);
         });
     }


nolimits4web:
But this issue is already been solved, it grab context (in case of it is absent) from cache here https://github.com/nolimits4web/Framework7/blob/master/src/js/router.js#L222

And i guess with your fix, if i want to call `.back()` with `force` and different custom context it will be replaced with context from cache

trancee:
I have tried it with your suggested fix, but if I do that, the context is not being applied to the second last template (and thus is empty and shows unexpected page once I navigate back there).

This only happens if I have more than 2 pages in the history, if I am on the third page, and then go back to the second page, the router will preprocess the first page during that step. And there is where the cached context is not applied to the template anymore. 

I agree that my fix did not account for the "force" option, but at least it is working as expected.

What do you suggest?

nolimits4web:
@trancee can you give me the minimal setup to replicate the issue and check for the correct fix? You can use JSFiddle/Codepen templates

trancee:
I have figured it out while writing the test case. I have the `template7Data` option set in the app configuration, so when I go back in the router, it is applying that one (which is empty) instead of using the correct data from the page itself.

I don't think that is correct behavior. As far as I understood, the `template7Data` should only be applied as a default if there is no context data given, or not?

Do you need me to send you my test case or did I explain it well enough for you to investigate this?

ZanderBrown:
You might want to fix the conflict

trancee:
@ZanderBrown I can't figure out what the conflict is about? There is no details about it. But we can close this pull request since the problem is actually somewhere else.

PR #1282, opened 2017-01-13
Version bumped to 1.5.3. Added isAnimated param to toolbar

centrual:
Hello Vladimir,

I added isAnimated parameter to:

- hideToolbar
- showToolbar
- hideNavbar
- showNavbar

methods. When you pressed an input in phone, animation looks bad. Because of that, i added isAnimated parameter. Default is true, so everybody can use same as before.

I wrote changes to CHANGELOG.md ( with your virtual list update )

Have a great day!


nolimits4web:
Thanks, i merged it manually because of the following reasons:
1. Version shouldn't be bumped
2. Too much indent space (4 spaces instead of 2 for tab)
3. I changed logic to contra verse. So if we don't pass new parameter, nothing should change, so no classes is added. And it adds "not-animated" class in case we pass "noAnimated" parameter

centrual:
Thank you Vladimir.

I will be careful.

Interestingly i using tab (2 spaces area) in webstorm but when i upload, it changes. I'll find why it is happening.

Have a great day!

PR #1286, opened 2017-01-14
Remove unused variables and redundant code in dom7-methods.js

blikblum:
This PR does

* Fix typo in comment (attriutes -> attributes)
* Remove redundant return in `prop` method
* Cache `this[0]` into `el` variable in `data` method
* Remove unused variable `j` in `transitionEnd` and `animationEnd` methods
* Remove unused variables and redundant return in `styles` method
* Cache `this[0]` into `el` variable and remove always false check (`el === document` will never evaluate to true when selector is a string)  in `is` method
* Initialize `child` variable to this[0] earlier and remove redundant return in `index` method
* Cache `this[0]` into `el` variable in `prev` method

### DIFF
--- a/src/js/dom7-methods.js
+++ b/src/js/dom7-methods.js
@@ -1,5 +1,5 @@
 Dom7.prototype = {
-    // Classes and attriutes
+    // Classes and attributes
     addClass: function (className) {
         if (typeof className === 'undefined') {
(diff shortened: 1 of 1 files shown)


nolimits4web:
Looks good, thanks!

PR #1287, opened 2017-01-14
Remove unused variables dom7-utils.js

blikblum:
This PR does:

* Fix typo in comment Utilites > Utilities
* Remove duplicate `url` variable declaration in `parseUrlQuery` (url parameter creates an implicit var)
* Remove unused variables doneLeft and doneTop in `scrollTo`

### DIFF
--- a/src/js/dom7-utils.js
+++ b/src/js/dom7-utils.js
@@ -1,6 +1,6 @@
-// DOM Library Utilites
+// DOM Library Utilities
 $.parseUrlQuery= function (url) {
-    var url = url || location.href;
+    url = url || location.href;
     var query = {},i, params, param, length;
     if (typeof url === 'string' && url.length)  {
         url = url.indexOf('?')>-1 ? url.replace(/\S*\?/,'') : '';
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks!

src/js/dom7-utils.js
```diff
@@ -1,6 +1,6 @@
-// DOM Library Utilites
+// DOM Library Utilities
 $.parseUrlQuery= function (url) {
-    var url = url || location.href;
+    url = url || location.href;
```
ZanderBrown: Woah now we are modifying parameters which isn't great.

How about `var toParse = url || location.href;` and continue with that?

blikblum: I am OK with that.

nolimits4web: @ZanderBrown `url` is a string so it won't be modified as it passed by value not by the link

blikblum: Anyway, if appropriate i can prepare a PR with the suggestion

PR #1291, opened 2017-01-15
Add support for a data-display-as attribute that specifies an alternate display value for select options

tbcrowe:
These changes add support for a data-display-as attribute that specifies an alternate display value for select options.  For example, when setting an alarm on iOS when the user sets the repeat days, the select shows values like "Every Monday", but on the main alarm page the value is displayed as "Mon".

A smart select set up to work similarly with these F7 changes might be defined as:

```html
    <li>
     <a href="#" class="item-link smart-select">
      <select name="repeat" multiple>
       <option value="0" data-display-as="Sun">Every Sunday</option>
       <option value="1" data-display-as="Mon">Every Monday</option>
       <option value="2" data-display-as="Tue">Every Tuesday</option>
       <option value="3" data-display-as="Wed">Every Wednesday</option>
       <option value="4" data-display-as="Thu">Every Thursday</option>
       <option value="5" data-display-as="Fri">Every Friday</option>
       <option value="6" data-display-as="Sat">Every Saturday</option>
      </select>
      <div class="item-content">
       <div class="item-inner">
        <div class="item-title label">Repeat</div>
        <div class="item-after">VALUE</div>
       </div>
      </div>
     </a>
    </li>
```

### DIFF
--- a/src/js/smart-select.js
+++ b/src/js/smart-select.js
@@ -23,7 +23,13 @@ app.initSmartSelects = function (pageContainer) {
 
         var valueText = [];
         for (var i = 0; i < select.length; i++) {
-            if (select[i].selected) valueText.push(select[i].textContent.trim());
+            if (select[i].selected) {
+                if (typeof select[i].dataset.displayAs !== 'undefined') {
+					valueText.push(select[i].dataset.displayAs.trim());
+				} else {
+					valueText.push(select[i].textContent.trim());
+				}
+			}
         }
 
         var itemAfter = smartSelect.find('.item-after');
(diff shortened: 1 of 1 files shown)


tbcrowe:
This is the first pull request I have ever submitted, so if I've missed a step or done something wrong please let me know.

tbcrowe:
Also, I have tested this with simple and multiple selects.  It works in both cases.

nolimits4web:
Thanks, merged! I also added additional check as `dataset` not supported everywhere

PR #1301, opened 2017-01-18
Allow to overwrite options in plugin hooks.

anatoly-mytrucking:
Call a plugin hook before extracting options into local variables.


nolimits4web:
Thanks! Merged

anatoly-mytrucking:
Thank you!

Issue #1288, opened 2017-01-14
Consider removing support for old browsers


blikblum:
This is a (multiple allowed):
* [ ] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)
* [x] question

Currently, F7 supports prefixed classes / events for old Opera (<= v10), IE (<= v9) and Firefox (<= v4) versions

Removing such support would help in file size as well runtime performance

As example, each `Dom7.transitionEnd` call listen to 5 events: ‘webkitTransitionEnd’, ‘transitionend’, ‘oTransitionEnd’, ‘MSTransitionEnd’, ‘msTransitionEnd’

By removing oTransition and MS(ms)TransitionEnd (IE10 supports transitionend) only 2 events would be created

Considering discussion in #241 and #308 the removal should not bring issues

Let me know if a PR removing such prefixes is appropriate

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/framework7)

Asked in the [forum](http://forum.framework7.io/#!/questions:what-browsers-are-supported), without response. Trying here

nolimits4web:
Agree, we can remove such prefixes. Feel free to send PR for this

valnub:
The PR was merged, so can this issue be closed?

PR #1303, opened 2017-01-19
Remove prefixed properties and methods for Opera, Firefox and IE

blikblum:
As discussed in #1288 

This PR does:
* Removes *Transform properties accesses
* Removes *TransitionDuration properties accesses
* Removes *TransitionEnd events
* Removes *AnimationEnd events
* Removes mozRequestAnimationFrame and mozCancelAnimationFrame

The non prefixed and webkit prefixed methods/properties are kept

It keeps `msMatchesSelector` since even IE 11 and EDGE support only the prefixed method

Tested in Firefox, Chrome, IE11, and everything seems to be working

### DIFF
--- a/src/js/dom7-methods.js
+++ b/src/js/dom7-methods.js
@@ -174,7 +174,7 @@ Dom7.prototype = {
     transform : function (transform) {
         for (var i = 0; i < this.length; i++) {
             var elStyle = this[i].style;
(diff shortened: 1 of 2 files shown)


nolimits4web:
Thanks, merged!

dalisoft:
But some users may want support older mobile browsers like Android 2.3.x or iOS 6. What they todo? Or this still supported?

nolimits4web:
@dalisoft these browsers are not supported by Framework7

dalisoft:
Can you show me list or graph of what browsers supported?

blikblum:
The webkit prefixes were kept so should have no problem

PR #1308, opened 2017-01-21
update .each() to emulate jQuery

### LINKED ISSUE #1305: [Bug] $$.each cannot break out of loop. DOM7
This is a (multiple allowed):
* [x] bug
* [x] enhancement

* Framework7 Version: 1.5.2
* Platform and Target: CHROME WEB BROWSER, CORDOVA-CROSSWALK, ANDROID.
* Live Link or JSFiddle/Codepen: https://codepen.io/anon/pen/xgqpmj

### What you did
DOM7 $$.each does not match expected behavior - how do you break out of the each statement in DOM7?
in JQuery if you return false it will break out but in DOM7 if you return false; it acts like continue; and moves onto next item in array.

### Expected Behavior
expect $$.each to break out of loop when return false callback is hit.

### Actual Behavior
$$.each when it hits return false .each just continues to next item in array - example in code pen above.
Expected to hit 'test' in array and break out of loop but you can see if just continues with the other items in the array

ZanderBrown:
Update DOM7's `each()` method to function similar to jQuery (https://api.jquery.com/each/)

Allows callbacks to return false to end the iteration of the given array/object

Should fix #1305

### DIFF
--- a/src/js/dom7-methods.js
+++ b/src/js/dom7-methods.js
@@ -423,10 +423,19 @@ Dom7.prototype = {
     },
 
     //Dom manipulation
+    // Iterate over the collection passing elements to `callback`
     each: function (callback) {
+        // Don't bother continuing without a callback
+        if (!callback) return this;
+        // Iterate over the current collection
         for (var i = 0; i < this.length; i++) {
-            callback.call(this[i], i, this[i]);
+            // If the callback returns false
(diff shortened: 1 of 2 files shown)


ZanderBrown:
@nolimits4web any thoughts?

nolimits4web:
Looks good, thanks!

PR #1320, opened 2017-01-25
Change .item-after class margin-left to auto, fixing list item right aligned content

### LINKED ISSUE #1314: Several bugs in Firefox 50
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Framework7 Version: 1.5.2
* Platform and Target: Android 6.0 via Firefox 50.1
* Live Link or JSFiddle/Codepen: PREFERABLY _(IF YOU WANT YOUR ISSUE TO BE RESOLVED ASAP)_. You can start from forking one of the templates: Material [Material Kitchensink](http://framework7.io/kitchen-sink-material/), iOS [iOS Kitchensink](http://framework7.io/kitchen-sink-ios/)

### What you did
Open the page. No coding modifications added.

### Expected Behavior
The page should render properly like in WebKit-based browser (Safari, Google Chrome, etc.)

### Actual Behavior
Bubble/Notification Tag (such as 'NEW' text on list) does not aligned properly, The bottom part of page is hidden by footer. This can be shown under the 'Messages' example.

### Further reason to file this bug
Some people are still using Mozilla Firefox, whether they are developing an app using Framework7 or visiting a mobile site powered by Framework7. As this issue happens in Android, I think that the issue also exist in desktop versions of Firefox. In case a regression in Firefox is confirmed, it may take months for the fix to be released, as it passes through many testing channels.

blikblum:
It changes margin-left to auto so list view items with right  aligned content is properly aligned in non webkit browsers. See rationale at https://medium.com/@samserif/flexbox-s-best-kept-secret-bd3d892826b6#.bi13yuk7i

Fixes #1314 

Tested with Firefox, Chrome, IE, EDGE

### DIFF
--- a/src/less/ios/lists.less
+++ b/src/less/ios/lists.less
@@ -111,7 +111,7 @@
         white-space: nowrap;
         color: #8e8e93;
         .flex-shrink(0);
-        margin-left: 5px;
+        margin-left: auto;
         .flexbox();
         max-height: 28px;
     }
--- a/src/less/ios/lists.less
+++ b/src/less/ios/lists.less
@@ -112,6 +112,7 @@
         color: #8e8e93;
         .flex-shrink(0);
         margin-left: auto;
+        padding-left: 5px;
         .flexbox();
         max-height: 28px;
     }
--- a/src/less/material/lists.less
+++ b/src/less/material/lists.less
@@ -109,7 +109,7 @@
         white-space: nowrap;
         color: #757575;
         .flex-shrink(0);
-        margin-left: 8px;
+        margin-left: auto;
         .flexbox();
         max-height: 28px;
         font-size: 14px;
--- a/src/less/material/lists.less
+++ b/src/less/material/lists.less
@@ -110,6 +110,7 @@
         color: #757575;
         .flex-shrink(0);
         margin-left: auto;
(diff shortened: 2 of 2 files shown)


nolimits4web:
When setting it to "auto" will remove 5px/8px spacing between title and it in case of long title or text in item-after which is not this case, e.g. in Media Lists. Probably it could be changed to padding, need to check won't it be a breaking change

blikblum:
I just checked and Media Lists will not be affected by this change since the margin-left property is kept due to rule of `.list-block .item-media + .item-inner`. I took screenshots and inspected in devtools to make sure.

In order to evaluate the change in an `.item-after` element i used the multiple values autocomplete example in kitchen sink and  selected 4 fruits.

In fact there's a difference but setting padding-left restored the previous appearance as show in the screenshots:

Current
f7-autocomplete-current-chrome

With margin-left: auto
f7-autocomplete-marginleftauto-chrome

With margin-left: auto and padding-left
f7-autocomplete-marginleftauto-paddingleft-chrome

blikblum:
I updated the PR setting also the padding-left

nolimits4web:
Ok now, thanks!

PR #1371, opened 2017-02-23
fix: bug in cbcdea4

jmarceli:
I think that there was a bug in commit cbcdea4. Variable `newValue[i]` is not an index of `values` column array, so it should be passed to `indexOf` in order to set correct display value of a picker


nolimits4web:
Merged, thanks!

PR #1372, opened 2017-02-24
Only include first level list elements in sortable

grgarside:
If an li in a sortable's ul contains li elements within, sortable tries to sort the parent li (or a different li entirely) within the child li. Fixes bug by only looking one level down for li elements.

### DIFF
--- a/src/js/framework7/sortable.js
+++ b/src/js/framework7/sortable.js
@@ -37,7 +37,7 @@ app.initSortable = function () {
         /*jshint validthis:true */
         sortingEl = $(this).parent();
         startIndex = sortingEl.index();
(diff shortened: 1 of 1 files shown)


grgarside:
It's possible I wasn't entirely clear (written quickly whilst I patched my production app!), so here's a gif which hopefully clarifies the problem. This is the problem which is fixed by my patch.

f7bugsortable2

ZanderBrown:
Looks good.
Just needs approval from @nolimits4web or @valnub

valnub:
I think it's better if @nolimits4web only merges PRs :)

nolimits4web:
Merged, thanks!

Issue #1364, opened 2017-02-16
Theme color is missing for `.statusbar-overlay`


iBinJubair:
This is a (multiple allowed):
* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* Framework7 Version: 1.5.3
* Platform and Target: iOS

### What you did
Adding class selectors lines to `framework7.ios.colors.css` and `framework7.material.colors.css` to cover theme colors. e.g.
`.theme-red .statusbar-overlay {
	background-color: #f44336;
}`

### Expected Behavior /  Actual Behavior
If this tag has been used `<meta name="apple-mobile-web-app-status-bar-style" content="black-translucent" />`, the status bar background color will be always `blue`. The `.statusbar-overlay` must be added to theme color css files.

valnub:
Is this a question or do you have a suggetion for improvement? If the latter is the case, you post your improvement as a pull request? Thanks

See https://help.github.com/articles/about-pull-requests/

macdonst:
Yeah, this looks like a good candidate for a PR. Also, whomever sends in a PR should make sure that they follow the the Material style guidelines. If the primary color is blue then the status bar should be primary color dark as per: https://developer.android.com/training/material/theme.html#StatusBar

macdonst:
I added a PR for this as it seemed like a fun way to spend a Sunday night learning .less.

nolimits4web:
Fixed in #1377 by @macdonst 👍

PR #1377, opened 2017-02-27
:sparkles: Issue #1364: Theme color is missing for .statusbar-overlay

macdonst:
The statusbar-overlay was stuck at blue even when you switch the theme color so I added in a rule for the statusbar-overlay class when the theme color changes. 

I also added in a bunch of dark colors for material as per the [Material color palette swatches](https://material.io/guidelines/style/color.html#color-color-palette). Using the 900 value for the dark color as the 500 value is the current theme color.

The dark color was added as per the [Material spec](https://developer.android.com/training/material/theme.html#StatusBar) as it specifies using the dark color for the statusbar overlay.

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -39,6 +39,7 @@
             "material" : [
                 "src/less/material/_mixins.less",
                 "src/less/material/_colors-vars.less",
+                "src/less/material/_dark-colors-vars.less",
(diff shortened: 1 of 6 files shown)


nolimits4web:
Great, thanks!

PR #1379, opened 2017-02-28
Change touch event listeners from passive to active to fix Chrome bug

brainformatik:
Passive touch event listeners have been changed from passive to active to fix mobile Chrome version >= 56 not preventing default.
Mobile Chrome version >= 56 makes event listeners passive by default.

We've detected the bug while using the sortable.

See:
[https://www.chromestatus.com/features/5093566007214080](https://www.chromestatus.com/features/5093566007214080)


J05HI:
I can confirm the bug. 
Chrome Mobile 57

nolimits4web:
Thanks for PR, but it is a bit wrong way to replace all to passive:false as we still need passive listeners in some places, so i've fixed it in a correct way

PR #751, opened 2015-12-10
Allow template7 page rerendering

antono:
Its possible to update template7 pages with new data.

PR #1391, opened 2017-03-03
Add page rerendering

J05HI:
This allows us to rerender the page with other data.
Helpful if the data has changed. With this you can easily rerender the page and musn't reinit the complete page nor change any data manually with .html()

See: #751 (I've updated it to be ready to merge)

### DIFF
--- a/examples/template7-pages/contacts-rerendered.html
+++ b/examples/template7-pages/contacts-rerendered.html
@@ -1,0 +1,19 @@
+<div class="navbar">
+  <div class="navbar-inner">
+    <div class="left sliding"><a href="#" class="back link"> <i class="icon icon-back"></i><span>Back</span></a></div>
+    <div class="center sliding">Contacts ReRendered</div>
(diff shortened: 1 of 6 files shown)


nolimits4web:
This PR has fundamental issue. As such rerendering will break all components initialized by "page:init" and won't detach all assigned events and destroy initialized components, so potentially can lead to memory leaks. So it is good for something simple but won't work and break things in case you have something more complex. Also, it won't work on iOS theme if you have some dynamic data in navbar because it is rendered outside of page. So to make it work correctly the page should be destroyed correctly then rerendered and added/replaced by new one and re-initialized. To make it work, it is possible to use view router's method, something like:
```js
function renderPage(data) {
    var tpl = Template7.cache[this.url];
    var newData = data || this.data || {};

    view.router.load({
        template: tpl,
        reload: true,
        reloadPrevious: $(this.container).hasClass('page-on-left'),
        context: newData
    });
}

J05HI:
Okay then I'll close it.

PR #1395, opened 2017-03-05
Add "clear N items", "select all" button to smart selects.

dourvaris:
This commit adds a button at the top right of a smart select which shows you how many items are selected, eg. "clear 5 items" and when no items are selected shows "select all"

This lets you select or deselect multiple items at the same time. 

Does not activate on:

- virtual lists
- pickers
- non multiple smart selects

If this feature is accepted, I'll update the docs as well.


dourvaris:
Has a bug on android due to preventDefault - closing until it's fixed.

PR #1405, opened 2017-03-10
Adjust modal horizontal position for non-fullpage application root

jmarceli:
While using `Framework7({root: '.some-selector' })` option and if your application container is not 100% wide the popover may be displayed outside the root container. Here is the screen (the application container is limited to max-width and centered on desktop browsers):
image

What do you think about fixing that issue? There is no problem in getting the following result:
image

In this PR the left offset and application width calculations are modified. Instead of absolute values `app.root` is taken into consideration. This is only a fix for modals other elements may still be affected.


J05HI:
Looks fine

nolimits4web:
Merged, thanks!

Issue #992, opened 2016-05-21
app.hidePreloader can close the wrong modal


plashenkov:
We’ve faced with an issue.

If several modals are displayed (for example, `app.alert()`, `app.showPreloader()`), even if “modal stack” is used, `app.hidePreloader` can close the wrong modal (in our example it closes alert instead of itself).

I assume we need to use a `.modal-preloader` class (instead of an abstract `.modal.modal-in` class) here: https://github.com/nolimits4web/Framework7/blob/master/dist/js/framework7.js#L3352

ZanderBrown:
@plashenkov is this still an issue?

plashenkov:
@ZanderBrown 
Well, original issue was created in May 2016. Looks like yes, it's still actual. See here: https://github.com/nolimits4web/Framework7/blob/master/dist/js/framework7.js#L3455

When we wrote an app we tried to use `app.hidePreloader()` and it closed the wrong modal for us. I suppose we have changed it to direct call to `app.closeModal('.modal-preloader')`.

plashenkov:
See here: https://github.com/nolimits4web/Framework7/blob/master/dist/js/framework7.js#L3451
`showPreloader ()` is created with `.modal-preloader` class. Why close it using generic class name `.modal.modal-in`? Is there any reason? This is why it closes wrong modal if two modals are displayed (we had exactly such case).

ZanderBrown:
Could you open a pull request?

macdonst:
@ZanderBrown I just opened a PR on it as it is such a small change.

https://github.com/nolimits4web/Framework7/pull/1410

plashenkov:
@macdonst Thank you!

nolimits4web:
Fixed by #1410

PR #1410, opened 2017-03-13
:bug: Issue #992: app.hidePreloader can close the wrong modal

macdonst:
Fixes issue #992. Closes preloader but not other modal dialogs if they are also open.


nolimits4web:
Thanks!

PR #1412, opened 2017-03-14
Allow User Specified Ratio for Lazy Load

### LINKED ISSUE #928: Lazy Load Allow User Specified Ratio
Hi,

Amazing framework by the way. Is it possible to make a very simple change to allow users to specify the initial src attribute on the image tags. I know there is imagesLazyLoadPlaceholder parameter but my page has varying image ratios which cause the page to jump when they load if they are not all a certain ratio. Some of my photos are 16x9 or 4x3 on one page etc. So I can't just set one placeholder image in imagesLazyLoadPlaceholder parameter.

In Framework7 1.4.2

If we change line 4295

`if ($(this).attr('data-src')) $(this).attr('src', placeholderSrc);`

to

`if ($(this).attr('data-src') && !$(this).attr('src')) $(this).attr('src', placeholderSrc);`

This allows users to simply specify the image tag as follows and scales the placeholder image correctly. Where lazy-16x9.jpg is 16x9 pixels in size and lazy-4x3.jpg is 4x3 pixels in size.

`
`

It would be great if we could get this added?

vestaxpdx:
This allows users to specify the initial src attribute on the image tags so they can choose different placeholder images that can vary in ratio. Rather than one global placeholder image.

Fixes #928

### DIFF
--- a/src/js/framework7/lazy-load.js
+++ b/src/js/framework7/lazy-load.js
@@ -33,7 +33,7 @@ app.initImagesLazyLoad = function (pageContainer) {
         placeholderSrc = app.params.imagesLazyLoadPlaceholder;
     }
     if (app.params.imagesLazyLoadPlaceholder !== false) lazyLoadImages.each(function(){
-        if ($(this).attr('data-src')) $(this).attr('src', placeholderSrc);
+        if ($(this).attr('data-src') && !$(this).attr('src')) $(this).attr('src', placeholderSrc);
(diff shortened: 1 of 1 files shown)


J05HI:
@nolimits4web 
Looks fine!

nolimits4web:
Merged, thanks!

vestaxpdx:
Cool!

PR #1416, opened 2017-03-15
Adds an onLazyLoad Callback

vestaxpdx:
It is sometimes really useful to detect when an image has been loaded via lazy load.

This adds this functionality.

```
var myApp = new Framework7({
     onLazyLoad: function (el) {
          console.log(el.attr('data-src'));
     }
});
```

### DIFF
--- a/src/js/framework7/lazy-load.js
+++ b/src/js/framework7/lazy-load.js
@@ -94,6 +94,9 @@ app.initImagesLazyLoad = function (pageContainer) {
         image.onload = onLoad;
         image.onerror = onError;
         image.src =src;
+		
+		// Add loaded callback
+		if (app.params.onLazyLoad && !el.hasClass('lazy-loaded')) app.params.onLazyLoad(el);
     }
     function lazyHandler() {
         lazyLoadImages = pageContainer.find('.lazy');


ZanderBrown:
As `$$('img').trigger('lazy');` manually triggers lazy load does `$$('img').on('lazy', function () {});` not work?

vestaxpdx:
I did think that but from testing the following just never gets called. Perhaps a timing thing?

```
var myApp = new Framework7({});

$$(document).on('lazy', 'img', function () {
    console.log('Image Loaded');
    console.log($$(this));
});
```

ZanderBrown:
Interesting. Still i think an event is the better way to handle this

vestaxpdx:
I am a fan of a callback function defined in the initialization. But I agree no need to edit the core if there is an event we can listen for. But I just can't get it to work.

vestaxpdx:
Correct me if I am wrong, but if I moved the code I added into onLoad() function it would fire when the image has actually fully loaded which is very beneficial in some cases. (slow connections)

Listening to the 'lazy' event would fire before the image has fully loaded?

nolimits4web:
Merged, thanks. Also added few events/callbacks at f78045edf01f012a511bf5275d6c0d440befd029

vestaxpdx:
Perfect cheers!

Issue #1089, opened 2016-08-10
Sortable List not working on Chrome / Windows


remko:
This is a (multiple allowed):
- [x] bug
- [ ] enhancement
- [ ] feature-discussion (RFC)
- Framework7 Version: 1.4.2
- Platform and Target: Chrome on Windows 8 (but reported on other windows versions as well)
### What you did

I tried dragging items from the sortable list on http://framework7.io/docs/sortable-list.html to reorder them.
### Expected Behavior

Items would be draggable
### Actual Behavior

Items do not move

I tried this on other browsers and platforms, and it works fine there.

ZanderBrown:
With mouse or touch?

remko:
I can only test it with mouse, but it has been reported with touch as well.

ZanderBrown:
ok because touch & mouse are handled differently. any code you can share?

remko:
I probably wasn't clear, but I meant to say that the example on http://framework7.io/docs/sortable-list.html exhibits the problem, so no need for custom code.

ZanderBrown:
In Chrome 54.0.2810.2 Windows 8.1 x86 I can confirm that sortable lists our unresponsive to mouse input but do respond to touch

Internet Explorer 11.0.9600.18427 Windows 8.1 x86 is responsive to mouse but not touch in both desktop & metro modes

*I can confirm this bug*

[Based on sample usage](http://framework7.io/docs-demos/sortable.html)

remko:
Why is the priority low on this one? This makes UIs that use listViews pretty much unusable on Windows. Is desktop not an important target for Framework7?

remko:
Since the PR has been reverted, can this issue be re-opened?

PR #1421, opened 2017-03-17
Fix touch detection on Windows browsers.

remko:
Chrome, IE, Firefox all have a 'null' 'ontouchstart' defined.
This broke touch detection, and hence things like listview reordering, swiping, ...

Fixes: #1089


nolimits4web:
Merged, thanks

nolimits4web:
Ok, i have to revert this fix as it is brakes touch detection where it is supported

PR #1435, opened 2017-03-21
Fix windows device detection error

brainformatik:
Fix error _"Unable to get property 'split' of undefined or null reference framework7.js (13461,17)"_ if using the device detection on windows phones.


nolimits4web:
Thanks

PR #1455, opened 2017-03-29
Fix suppressed errors for code inside ajax

brainformatik:
If wrong code is in an ajax call, it has been suppressed by Framework7.
With this fix it throws the error in the console.


J05HI:
@nolimits4web @valnub @ZanderBrown 
Tested and it's fixed now. It's strange no one has reported that before.

ZanderBrown:
That's because you should listen for the `ajax:error` event.

Or if using `$.ajax` set the `error` or `statusCode` argument

J05HI:
@ZanderBrown 
No its not the ajax:error event. 
It's when the code inside the success function has an error.

ZanderBrown:
in that case the callback itself should contain a try { } catch () { }

brainformatik:
@ZanderBrown 
Okay we'll use try catch, thank you!

PR #1472, opened 2017-04-04
fix swiper in module.json for custom build

bingzheyuan:
swiper js path in module.json is wrong, swiper is not included in custom build


ZanderBrown:
`swiper-init.js` probably shouldn't be in `framework7` and ideally the folder should be called `swiper7` not `swiper`

nolimits4web:
Thanks

PR #1517, opened 2017-04-23
toggle `focus-state` and `not-empty-state` in material input form

achmadk:
**Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.**
According to [this issue](https://github.com/nolimits4web/Framework7/issues/1516), the link below input value is not disappear because `focus-state` still exist although `not-empty-state` has been added to the input. So I remove `not-empty-state` class when input is on focus and remove `focus-state` when input is on blur. And then I add css styles of `.${STATE}.item-inner label, .${STATE}.input-field:after,  .${STATE} .item-input-field:after` with STATE value is `focus-state` and `not-empty-state`

### DIFF
--- a/src/js/framework7/material-inputs.js
+++ b/src/js/framework7/material-inputs.js
@@ -31,7 +31,7 @@ app.initMaterialWatchInputs = function () {
         var type = i.attr('type');
         if (notInputs.indexOf(type) >= 0) return;
         var els = i.add(i.parents('.item-input, .input-field')).add(i.parents('.item-inner').eq(0));
-        els.addClass('focus-state');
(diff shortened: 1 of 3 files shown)


J05HI:
I think it's fine, isn't it?

ZanderBrown:
@J05HI we can squash and merge on GitHub. Thank you for seconding the PR.

Now we
```js
let canMerge = await nolimits4webReview():
```

nolimits4web:
Looks good, thanks!

PR #1527, opened 2017-04-25
FIX: if previousQuery doesn't exist,but searchterm

Phillipip:

https://github.com/nolimits4web/Framework7/issues/1526

Always follow the [contribution guidelines](https://github.com/nolimits4web/Framework7/blob/master/.github/CONTRIBUTING.md) when submitting a pull request.


ZanderBrown:
Doesn't look like it'll cause any issues but i'm not sure why you are doing this.

nolimits4web:
Looks ok to me

PR #1547, opened 2017-05-07
black image fix

scriptPilot:
replaced black image


ZanderBrown:
What? Why?

scriptPilot:
I replaced a URL as there is no image behind anymore. Why? Color is more beautiful than black I would say :-)

ZanderBrown:
Oh you fixed a broken URL?

nolimits4web:
Thanks ;)

PR #1552, opened 2017-05-10
secure all urls

scriptPilot:
All urls checked before if SSL is possible. This will avoid warnings in
some web browsers!


ZanderBrown:
How about `//`ing all the things

scriptPilot:
@ZanderBrown - What do you mean?

ZanderBrown:
[`//` Magic](https://www.paulirish.com/2010/the-protocol-relative-url/)

scriptPilot:
@ZanderBrown - Thanks, I updated all src files accordingly!

valnub:
Well done, how can we check if all url's have been changed and nothing broke?

valnub:
Hold on a sec, the changes seem to be made for xmlns:

`<svg xmlns='//w3.org/2000/svg'`

Is this really necessary? I don't think that the browser actually calls these url's. They're just namespaces.

nolimits4web:
Closing as incompatible with current version and probably not needed anymore

PR #1554, opened 2017-05-11
show element without affecting on the display style of it

WangHaoming:


WangHaoming:
This fix will effect on popup function which call show function for displaying popups. need to find a better way to fix it.

nolimits4web:
Closing as incompatible with current version and probably not needed anymore (was fixed in Dom7)

PR #1566, opened 2017-05-22
Autocomplete behavior error

spltfc:
Should insert value instead of label to input element.
(maybe can fix this issue too https://github.com/nolimits4web/Framework7/issues/934)


nolimits4web:
Looks good, thanks!

PR #1620, opened 2017-06-01
Ported Cards from v1 to v2 | closes #1584

### LINKED ISSUE #1584: [v2] Cards Component
Depends on #1602

valnub:
Note:

.button undefined in card-md.less

Replaced with:

    text-transform: uppercase;
    padding-left: 8px;
    padding-right: 8px;

Maybe change later when .button is available

### DIFF
--- a/pages/cards.html
+++ b/pages/cards.html
@@ -8,6 +8,189 @@
     </div>
   </div>
   <div class="page-content">
-    <p>Cards goes here</p>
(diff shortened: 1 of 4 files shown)


nolimits4web:
Merged, thanks! Looks great, excepts -ios/-md styles should be nested under html.ios, html.md rules, otherwise they will overwrite each other. I've quickly fixed it

Issue #1602, opened 2017-06-01
[v2] Button Component


nolimits4web:

ZanderBrown:
Work-in-progress https://github.com/ZanderBrown/Framework7/tree/v2

J05HI:
@nolimits4web 
What about adding this type of buttons?

![screenshot_20170624-145444](https://user-images.githubusercontent.com/16399787/27508821-f5335caa-58ed-11e7-86d0-558f9e39a732.png)

J05HI:
@nolimits4web 
Great! Thank you very much :)

numberokta:
When touch and hold two button together ripple stuck on one of them !

PR #1623, opened 2017-06-02
V2 Buttons

ZanderBrown:
Simple implementation of buttons

Probably imperfect

Stray thought:
Possibly use autoprefixer?


ZanderBrown:
**This is work in progress**

ZanderBrown:
#1602

nolimits4web:
I see you have a bit wrong color-vars reference, -iOS, -md colors should be imported under HTML.ios, HTML.md rules in relevant files

nolimits4web:
Merged, thanks!

PR #1634, opened 2017-06-08
fix picker selecting incorrect item on scroll up

johnzondr:
We are noticing that when we scroll up in the picker, and we scroll between two times, then the new time is selected, but when you remove your fingers then the the picker selects the old time, and not the new time. 

[Gif of bug](https://gyazo.com/399d3be580bbd7a40baad0a1a654b029)


johnzondr:
@nolimits4web please take a look

nolimits4web:
Closing as incompatible with current version and probably not needed anymore. If issue is still exist in v2 then send new PR for the v2

PR #1641, opened 2017-06-13
Propose a fix on smartSelectAddOption function

ismaelgsan:
When you need to open a page with smart select showed twice, the function "smartSelectAddOption" open the previous smart select too.
This fix open the last smart select only.

Thank you.


nolimits4web:

Issue #1662, opened 2017-06-26
Stacked modals not working correctly


adrian-e:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Framework7 Version: 1.6.4
* Platform and Target: All platforms
* Live Link or JSFiddle/Codepen: https://codepen.io/anon/pen/bRoZyo

### What you did
Created a modal and then trigger an alert from inside said modal (it's a validation prompt in the real app).

### Expected Behavior
I would expect the alert to dim the modal content and when it is closed to still have the overlay of the modal to prevent the user from interacting with the rest of the app

### Actual Behavior
The alert does not dim the modal that prompted it, and when the alert is closed the modal's overlay is also closed.

nolimits4web:
Issue is closed because of outdated/irrelevant/not actual

If this issue is still actual and reproducible for latest version of Framework7, please create new issue and fill the issue template correctly:
* Clearly describe the issue including steps to reproduce when it is a bug.
* Make sure you fill in the earliest version that you know has the issue.
* Provide live link or JSFiddle/Codepen or website with issue

PR #1663, opened 2017-06-27
Stacked modals not working correctly

adrian-e:
- Make sure that each stacked modal is shown above the previous one
- Create an overlay for each stacked modal
- After close of each modal remove corresponding overlay without affecting others
refs #1662 

For me this solved the issue and allows multiple stacked modals (I have used modal and alert only).


nolimits4web:

PR #1673, opened 2017-07-03
Added attribute `data-close-manual` for  `.accordion-item` node. 

frontenddeveloping:
if `data-close-manual` is present on`.accordion-item` item can be closed only by direct click on it'.
This changes provides ability to expand all items and close only clicking on each item.


frontenddeveloping:
@nolimits4web ping

nolimits4web:
Isn't it the same if you just remove "accordion-list" class from "list-block"? In this case opening one item won't close another one

nolimits4web:

PR #1680, opened 2017-07-07
Bug fix: Picker

vulvulune:
When there are dividers columns, the displayValues are incorrect in the formatValue callback on the first load when an initial value is specified.

The bug with the version 1.6.3 (without this PR):
https://jsfiddle.net/Ljntx07v/4/


nolimits4web:
Merged, thanks

PR #1688, opened 2017-07-16
Fix check to current page at view initialization

### LINKED ISSUE #1687: View history is wrong when a view is initialized with a empty pages element
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Framework7 Version: 1.6.4
* Platform and Target:  web

### What you did
Initialized a view with an empty pages element

### Expected Behavior
No page should be cached or added to view history

### Actual Behavior
A page with name #undefined is added to view history

blikblum:
Fix #1687

When pages element is empty currentPage variable is truthy but its length is 0. This is set view.url to '#undefined' and mangles the view page cache and history

When currentPageData is truthy means that currentPage is also truthy with length === 1, so a proper url can be set

### DIFF
--- a/src/js/framework7/views.js
+++ b/src/js/framework7/views.js
@@ -122,7 +122,7 @@ var View = function (selector, params) {
     }
 
     // View startup URL
-    if (view.params.domCache && currentPage) {
+    if (view.params.domCache && currentPageData) {
         view.url = container.attr('data-url') || view.params.url || '#' + currentPage.attr('data-page');
         view.pagesCache[view.url] = currentPage.attr('data-page');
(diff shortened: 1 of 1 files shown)


src/js/framework7/views.js
```diff
@@ -122,7 +122,7 @@ var View = function (selector, params) {
     }
 
     // View startup URL
-    if (view.params.domCache && currentPage) {
+    if (view.params.domCache && currentPageData) {
```
blikblum: It can be also `view.params.domCache && currentPage && currentPage.length`

PR #1748, opened 2017-08-21
Replace + with spaces in queries

### LINKED ISSUE #1747: pages.query does not decode '+' spaces
This is a (multiple allowed):
* [x] bug

* Framework7 Version: 1.6.4
* Platform and Target: IOS, ANDROID

### What you did
```javascript
// /#!/error.html?msg=There+was+a+problem
myApp.onPageInit('error', function (page) {
    $$('#error_msg').text(page.query['msg']);
});
```

See [this Stack Overflow comment](https://stackoverflow.com/questions/1005676/urls-and-plus-signs#comment14903296_1006074) for a discussion about the "standards-ness" of the plus sign in queries.

### Expected Behavior
`page.query['msg']` should contain `"There was a problem"`

### Actual Behavior
`page.query['msg']` contains `"There+was+a+problem"`

eslindsey:
Fixes #1747

### DIFF
--- a/src/js/dom7/dom7.js
+++ b/src/js/dom7/dom7.js
@@ -175,7 +175,7 @@ var Utils = {
 
       for (i = 0; i < length; i += 1) {
         param = params[i].replace(/#\S+/g, '').split('=');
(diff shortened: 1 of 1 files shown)


nolimits4web:

PR #1749, opened 2017-08-21
Run <script>s in loaded pages

### LINKED ISSUE #70: Trig Javascript in .html file after loading page.
Hello!
When you click on a link, an Ajax call is done and the page is loaded. But in this page, there is Javascript content that can trigger external libraries' content like Highcharts for example.

Are there any way to use Javascript in .html file after an Ajax call loading page ?

Thanks for your Framework!
CloudCompany

eslindsey:
Enhancement that fixes #70

### DIFF
--- a/src/js/framework7/pages.js
+++ b/src/js/framework7/pages.js
@@ -65,6 +65,25 @@ app.triggerPageCallbacks = function (callbackName, pageName, pageData) {
             allPagesCallbacks[j](pageData);
         }
     }
+
+    // Run scripts
+    $(pageContainer).find('script').each(function() {
+        // Check for data-run
+        var wantedCb = $(this).data('run').toLowerCase() || 'init';
+        if (callbackName.toLowerCase() !== wantedCb) {
+            return;
+        }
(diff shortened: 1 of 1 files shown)


eslindsey:
This enhancement now allows an optional, case-insensitive `data-run` attribute on the `SCRIPT` tag, which determines when that script is executed. All page callback values are allowed: `beforeInit`, `init`, `reinit`, `beforeAnimation`, `afterAnimation`, `back`, `afterBack`, `beforeRemove`. If `data-run` is not provided, it defaults to `init`.

eslindsey:
Note that defining functions within inline scripts executed in this method does not work quite as expected. Instead of writing:
```javascript
function myFunctionHere() {
    //...
}
```
You must define the function in the style of a variable:
```javascript
window.myFunctionHere = function() {
    //...
}
```

ZanderBrown:
Not sure about the use of eval, perhaps use new Function or a data uri?

lbrawnbrawny:
good

nolimits4web:

PR #1751, opened 2017-08-21
Dark Blue Layout As “.layout-dark-blue”

Banakin:
New iOS dark blue layout as “.layout-dark-blue” based off of the dark layout “.layout-dark”


docincmh1:
Please help me fix my Android with thank you to

nolimits4web:
Closing as incompatible with current version and probably not needed anymore. If you think this is still the case in v2 then make PR for v2.

Also, themes are not often used by everyone and brings a lot of extra CSS. So probably it will be better to introduce it as a plugin

PR #1761, opened 2017-08-27
ported notifications from v1

kozmanbalint:
I'm slowly moving towards v2 with my apps, and notifications are a must in my case.
To resolve this I've ported notifications from v1, which at least means they use css transitions, etc. Adjusted styles and the api to make them work with v2 while tried to keep it as much backward compatible as possible.

Cheers,
Bálint

### DIFF
--- a/kitchen-sink/index.html
+++ b/kitchen-sink/index.html
@@ -186,7 +186,7 @@
                 <a href="/navbar/">Navbar</a>
               </li>
               <li>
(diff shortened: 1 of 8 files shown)


nolimits4web:
Thanks! Merged for now. But i'm thinking about to rework a bit its API with adding support of Push Notifications as well. Will do it in a close feature. Great for now

kozmanbalint:
Thanks.

PR #1764, opened 2017-08-28
Fix issue with wrong element selected when scrolling

bencompton:
In iOS home screen apps, when momentum scrolling is started and then an element is tapped while the element is still scrolling, the wrong element will be selected. This only happens with FastClicks enabled.

To reproduce, add the kitchen sink to home screen on an iOS device, start the main menu page scrolling, then tap on a link. Notice that you end up on a different page than the link you tapped.

Note that this fix does not ensure that `active-state` gets set on the correct element, but it does ensure that the click event gets fired for the right element. So for example, the you will now always be taken to the correct page in the kitchen sink, but the wrong link may be highlighted. Not perfect, but it is definitely an improvement.


nolimits4web:
Thanks!

PR #1765, opened 2017-08-28
[V2] Fix issue with the wrong element selected when scrolling

bencompton:
Same fix as #1764, but for V2.


nolimits4web:
Merged, thanks!

PR #1773, opened 2017-09-02
fix md colors generation

kozmanbalint:
build script typofix


kozmanbalint:
Has been fixed in line with the upstream - not relevant anymore.

PR #1780, opened 2017-09-07
[v2] Add check to $pageEl in pageCallback

blikblum:
It fixes a bug when there are currently no page in the view. POC at https://codepen.io/blikblum/pen/MvdrWO?editors=1011

When there's no page, the $oldPage is empty and when passed as argument to pageCallback ('beforeOut' callback) an exception occurs


nolimits4web:
Good thanks!

PR #1782, opened 2017-09-08
Cleanup pageCallback: remove double check for callback argument

blikblum:
This PR removes double check for callback argument (=== 'beforeRemove') in Router.pageCallback

Also remove setting page to null since, page is an internal variable not used elsewhere


nolimits4web:
Thanks

PR #1783, opened 2017-09-08
[v2] Use native Promise instead of custom implementation

blikblum:
This PR replaces the custom implementation by a call to the native Promise.

The advantages are less code, inherits browser [performance improvements](https://v8project.blogspot.com.br/2017/02/v8-release-57.html), ensure Promise A+ compliance

The only drawback is that the developer targeting browsers that do not support Promise will need to provide a Promise polyfill.

iOS safari supports Promise since v 8.0 released september 2014.
Chrome since v 33 and Firefox v 29. IE does not supports at all but given the F7 scope i don't see as a problem.

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -98,6 +98,49 @@ for (let i = 0; i < defaultDiacriticsRemovalap.length; i += 1) {
   }
 }
 
+const createPromise = function(handler) {
+  let resolved = false;
+  let rejected = false;
+  let resolveArgs;
+  let rejectArgs;
+  const promiseHandlers = {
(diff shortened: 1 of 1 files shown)


nolimits4web:
I agree to use native Promises but that polyfil should be still there as IE i still a target because F7 also used as desktop apps. We can just add `if (window.Promise) return new window.Promise(handler)` in the beginning of this util

blikblum:
> I agree to use native Promises but that polyfil should be still there

Done. I updated the PR accordingly

nolimits4web:
Thanks

PR #1809, opened 2017-09-20
[v2] minor fixes

pi0:
Hi. This PR fixes rollup build warning messages by explicitly define external and globals for `dom7` and `template7`. 

Also in order make empty `pushStateSeperator` support, we need an extra check to make sure not split `documentUrl` using empty string which makes `initUrl` undefined and cusing errors. This change has been tested to working with latest f7-vue commit.

### DIFF
--- a/scripts/build-js.js
+++ b/scripts/build-js.js
@@ -23,6 +23,16 @@ const banner = require('./banner.js');
 
 let cache;
 
+const external = [
+  'dom7',
+  'template7'
+]
+
+const globals = {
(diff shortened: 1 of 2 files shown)


nolimits4web:
Thanks!

pi0:
Thanks for merges @nolimits4web :) With this one, now [Nuxt7](https://github.com/pi0/nuxt7) is basically working! Many many improvements and changes are needed. I'll keep it updated with F7 :)

PR #1819, opened 2017-09-22
[v2] fix: assign swiper on initSwipers

pi0:
This PR fixes problems with swipeable tabs which need `.swiper` reference.  ([here](https://github.com/framework7io/Framework7/blob/v2/src/components/tabs/tabs.js#L68)) Also, `swiper` existence is ensured to prevent any uncaught errors.

### DIFF
--- a/src/components/swiper/swiper.js
+++ b/src/components/swiper/swiper.js
@@ -56,7 +56,7 @@ function initSwipers(swiperEl) {
     });
   }
 
-  $swiperEl[0].swiper = app.swiper.create($swiperEl[0], params);
+  app.swiper.create($swiperEl[0], params);
 }
 
 export default {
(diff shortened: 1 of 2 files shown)


nolimits4web:
I agree with having double check for swiper, but `swiper` element assignment already happens in the swiper core at https://github.com/framework7io/Framework7/blob/v2/src/components/swiper/swiper-class/components/core/core-class.js#L89

pi0:
Yes. So that may be another problem but if we combine (swiperable + routable) tabs example and without this line it throws an error that `swiper` is undefined. BTW I can revert 9e8c18d as you already marked that demos as TODO :)

nolimits4web:
Thanks! Yes, there is an issue with swipeable-routable tabs. They don't support to be "routable" at the moment.

PR #1838, opened 2017-09-28
refactor: bring component prior to name when loading route

clarkdo:
When using [nuxt7](https://github.com/pi0/nuxt7) which is based on [Framework7-Vue V2](https://github.com/framework7io/Framework7-vue), will get error with ios theme.

Because we defined `name` on every `routes` for some further use and [navigate](https://github.com/framework7io/Framework7/blob/v2/src/modules/router/navigate.js#L464-L468) will take take precedence over `component` and not call `pageComponentLoader`, so codes inside `forward` method will throw error due to empty component html.

The original bug description is in https://github.com/pi0/nuxt7/pull/8. Pls feel free to discuss with me to figure out a appropriate solution.

### DIFF
--- a/src/modules/router/back.js
+++ b/src/modules/router/back.js
@@ -464,7 +464,7 @@ function back(...args) {
     });
   }
 
-  ('url content name el component componentUrl template templateUrl').split(' ').forEach((pageLoadProp) => {
+  ('url content component name el componentUrl template templateUrl').split(' ').forEach((pageLoadProp) => {
     if (route.route[pageLoadProp]) {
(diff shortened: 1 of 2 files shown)


nolimits4web:
Good thanks! The issue here is because `name` in f7 refers to page name in DOM (for pages with `data-name` attribute), not to route name. But i think i can make `name` to point route name to create named routes, and change to `pageName` what is now called `name`

clarkdo:
@nolimits4web  thanks for your patient reply. If would be fantastic if `name` can have a refactor.

PR #1842, opened 2017-09-30
[v2] viewmodel binding for components

kozmanbalint:
Hi Vladimir, I've just added basic one-way viewmodel binding feature to f7 component.
This is much rather a tech experiment and maybe not the right moment in beta8 to add features like this, but probably worth a shot.
The basic idea is to replace component data (the viewmodel, actually) with a proxy on component creation, and catch all changes in proxy's set. Upon real viewmodel changes the new component html is created (using template7 and the current viewmodel data) and the dom is patched to the new state with morphdom. Finally all event handlers are reattached, etc.
The implementation is far from being complete, however things basically seem to work.

Sorry for the commit storm, obviously only 795150730c49bebf16f7e0f5e2e582529593c9d3 and 499b2550b7abf4fe5f9590e44c58a07ac80b30f9 are relevant (is it possible to omit the rest from the pr? :-o).


J05HI:
Looks very interesting :)

D-Marc1:
Is this a two-way data binding solution? If so, that's awesome.

J05HI:
@D-Marc1
He wrote its basic one-way data binding

D-Marc1:
Then I really don't understand what this is, as Template7 currently already is a one-way data binding library

kozmanbalint:
Well, I'm maybe mistaken, but Template7 is a templating engine - and as such it _will not_ watch for datacontext changes and re-render parts of the DOM that have actually changed. This behaviour is what this pr is about.

'One-way' means that any change in the datacontext is reflected in the DOM, however direct DOM changes will not affect the datacontext (to avoid circular references, etc).

J05HI:
That's great
No jquery dom changes anymore...

D-Marc1:
@kozmanbalint thanks for clearing that up. That's pretty cool.

kozmanbalint:
@J05HI Avoiding direct dom7 manipulation on datacontext (aka. viewmodel) changes: that's the name of the game. Without that one will duplicate presentation logic - to say at least. And it will spread among the template source and the eventhandlers, which hardly produces easily maintainable code.
However the above implementation is far from being complete, it's just that I was curious what Vladimir and the community thinks about the concept.

nolimits4web:
Sorry it was closed automatically with the merge of v2 branch to master. I like the idea and will explore this after i finish v2 docs and can focus on such improvement

kozmanbalint:
No problem. I also have some more thoughts on the implementation, eg. suspend/resume layout to perform batch updates effectively on the viewmodel, etc. Will remove my fork and submit a new pr later.

nolimits4web:
Also probably maybe an issue here when we have a rerendering is that auto initialized components (what F7 does on page init). If such components added dynamically then init/reinit will be required

ffahd:
F

ffahd:
Cgcsdgg

PR #1847, opened 2017-10-04
Add data-table to list of available custom modules

blikblum:
`data-table` component is not available in custom module list. This PR adds it.


nolimits4web:
Thanks

PR #1850, opened 2017-10-05
Target .tablet-auto siblings instead of .col-auto

telwing:
The .tablet-auto class was applying each incremental :nth-last-child width respectively to .tablet-auto elements instead of consistently using the smallest :nth-last-child width.


nolimits4web:
Thanks

Issue #2044, opened 2018-01-04
theme switch broken in cordova app


yelban:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Framework7 Version: 2.0.5
* Platform and Target: Cordova android/ios

### What you did
switch theme beside kitchen-sink/index.html, got "Application Error" (Android) or nothing happen (ios)

### Expected Behavior
switch to ios theme (or material theme)

### Actual Behavior
got "Application Error" (Android) or nothing happen (ios)

![a77be7](https://user-images.githubusercontent.com/981381/34555426-aea10b44-f16c-11e7-8372-081b4c07bf96.png)

yelban:
submit a PR to fix this issue.
[PR #2045 ](`https://github.com/framework7io/framework7/pull/2045`)

nolimits4web:
Fixed by #2045, thanks

PR #2045, opened 2018-01-04
fix theme switch broken in cordova app

yelban:
this PR fix theme switch in cordova app as [#2044](https://github.com/framework7io/framework7/issues/2044)


nolimits4web:
Thanks

PR #2062, opened 2018-01-07
correct formula in pull-to-refresh-class.js

nimo23:
I changed "touchesDiff ** 0.85" to "touchesDiff * 0.85".


nimo23:
Oh, "**" is the ES6-"Exponentiation assignment", so it was correct.

PR #2114, opened 2018-01-17
fix closeByBackdropClick not working

wenjiamian:
fix closeByBackdropClick not working


nolimits4web:
Thanks, but such logic must be in popover-class, so i moved it there

PR #2161, opened 2018-01-30
Remove redundant code from router-class.js

blikblum:
Removes redundant code:

 * `return undefined` in end of `findElement`. When there's no return statement, undefined is returned
 * remove undefined after return since it will also return undefined when no arg is used
 * assignment to null for $pageEl properties in `detachEvents`. Those properties will be deleted just after
 * assignment to null for router var in destroy. Variable not used after


nolimits4web:
Sorry, but project uses ESLint and its "consistent-return" flag, https://eslint.org/docs/rules/consistent-return

About null, it is more about memory management, and considered good pratice to "null" things just to make sure.

Closing for now, as it doesn't actually improve or fix anything

PR #2183, opened 2018-02-06
Fix Action Sheet safe area on iPhone X

TheJKM:
This fixes a small typo related to iPhone X support and the action sheet component.


nolimits4web:
Thanks

PR #2185, opened 2018-02-07
Remove ios theme restriction for handling `no-navbar` and `no-toolbar` CSS classes. (#2184)

### LINKED ISSUE #2184: [v2] `no-navbar` and `no-toolbar` CSS class not working in material design (md) theme
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* Framework7 Version: framework7@2.0.6
* Platform and Target: tested in Chrome, target Cordova (Android).
* Live Link or JSFiddle: https://jsfiddle.net/w69Le27q/1/

### What you did
I try to hide navbar and toolbar in the About page (cf. https://jsfiddle.net/w69Le27q/1/) with CSS class `no-navbar` and `no-toolbar` in a `md` themed app.

### Expected Behavior
Both navbar and toolbar should be hidden.

### Actual Behavior
They are both visible.

---

In the JSFiddle if you change `theme: 'md'` to `theme: 'ios'` you can see that both navbar and toolbar are hidden in the About page.

Indeed:
https://github.com/framework7io/framework7/blob/621d71562af0f63e661a4ea327ad8544c83ccc8d/src/components/navbar/navbar.js#L271-L288

Why this line?
https://github.com/framework7io/framework7/blob/621d71562af0f63e661a4ea327ad8544c83ccc8d/src/components/navbar/navbar.js#L273

My feeling is that this little bug could be resolved by removing this line.

Same for toolbar:
https://github.com/framework7io/framework7/blob/621d71562af0f63e661a4ea327ad8544c83ccc8d/src/components/toolbar/toolbar.js#L150

Jule-:
If there is no particular reason to not handle these classes in Material Design theme, I think theses two changes can't be harmful and fix #2184.

### DIFF
--- a/src/components/navbar/navbar.js
+++ b/src/components/navbar/navbar.js
@@ -270,7 +270,6 @@ export default {
     },
     pageBeforeIn(page) {
       const app = this;
-      if (app.theme !== 'ios') return;
       let $navbarEl;
       const view = page.$el.parents('.view')[0].f7View;
       const navbarInnerEl = app.navbar.getElByPage(page);
--- a/src/components/toolbar/toolbar.js
+++ b/src/components/toolbar/toolbar.js
@@ -147,7 +147,6 @@ export default {
     },
     pageBeforeIn(page) {
       const app = this;
-      if (app.theme !== 'ios') return;
       let $toolbarEl = page.$el.parents('.view').children('.toolbar');
       if ($toolbarEl.length === 0) {
         $toolbarEl = page.$el.find('.toolbar');


nolimits4web:
Thanks, i merged it manually and at the moment for Toolbar only. The thing with navbar is basically for iOS theme only for dynamic navbar. In MD navbar is in the page, so if you don't need the navbar on page you basically set `no-navbar` (will hide common navbar in iOS theme) and remove navbar element from page

Jule-:
Thanks for the Toolbar manual merge! 🙂 

Just to be clear,
- You are suggesting that for iOS and MD we have to construct F7 views differently?

or

- You mean that F7 is taking care of switching `<div class="navbar">` from `<div class="page">` to `<div class="view">` depending on `theme` and `iosDynamicNavbar`?

or

- You mean that we need to set both navbars?
In MD the `.view`**`.navbar`** is hidden since it is the iOS specific dynamic navbar and in iOS the `.view .page`**`.navbar`** is hidden by setting the `no-navbar` class?

Thanks for clarification.

nolimits4web:
F7 is taking care of switching `<div class="navbar">` from `<div class="page">` to `<div class="view">` depending on theme and iosDynamicNavbar?

Issue #2200, opened 2018-02-13
V2 router not able to navigate/back after navigated to a page with async second time


robinyip:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Framework7 Version: 2.0.7
* Platform and Target: all

### What you did
routes: [
  {
    path: '/page',
    async: function( routeTo, routeFrom, resolve, reject){
      .................
    }
  {
]
view.router.navigate( '/page');     <--- first load
view.router.navigate( '/page');     <--- second load
after second load, you can't back or navigate to any page without async()

### Expected Behavior
Should be able to navigate/back

### Actual Behavior
Back or navigate to any page without async() will be failed

PR #2202, opened 2018-02-14
Update framework7.js

robinyip:
Fixed the issue #2200

This fixed the unable to navigate/back problem when a page has been called twice with async function defined in route


nolimits4web:
Thanks, i have merged it manually as changes must be pushed only to /src/ files

PR #2215, opened 2018-02-20
Add a difference on "distance from side of screen for slide panels swipe" when the slide is open

SpiderNeo1:
Hello,
I am in French so forgive me for my writing.

When you choose to open the panel on the right, there is the property "swipeActiveArea" which allows to define the distance from which we will be able to open the slide. It's very good.

But it misses the same thing when the panel is open on the right on the open area to tell from what distance I can close the panel.
This is important because when we have on the panel elements that we can "swipe to delete", it will close the panel.

If you want, I can make a video to show you.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -10,6 +10,7 @@ export default {
       rightBreakpoint: 0,
       swipe: undefined, // or 'left' or 'right' or 'both'
       swipeActiveArea: 0,
+      swipeActiveAreaSide: 0,
       swipeCloseOpposite: true,
       swipeOnlyClose: false,
       swipeNoFollow: false,
--- a/src/components/panel/swipe-panel.js
+++ b/src/components/panel/swipe-panel.js
@@ -53,10 +53,10 @@ function swipePanel(panel) {
       }
     }
     if (params.swipeActiveAreaSide && panel.opened) {
-		  if (side === 'right') {
-		    if (touchesStart.x > (app.width - $el.outerWidth() + params.swipeActiveAreaSide)) return;
-		  }
-	  }    
+      if (side === 'right') {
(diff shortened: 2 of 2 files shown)


SpiderNeo1:
What is the problem with "The Travis CI build" ?
Someone can help me please ?

nolimits4web:
Well, this is not really the solution to integrate it because it has conflicts with swipeouts. In this case it just needs to be fixed in another way like if there is a swipeout then don't close it. Also, even if go with such new option feature then apparently it must work for both panels not for right only

SpiderNeo1:
Hello Vladimir,
Thank you for your answer.

This is not only for the "swipe to delete" but it can be annoying for users (on some applications that have a panel with scroll) to see the panel close unintentionally because of a too large area.

You had understood that when you created the "swipeActiveArea" property.

I remain convinced that it is a good thing for F7 to add it.

And if you have to code the left side, I can do it so that it works for both panels.

Are you ok ?

nolimits4web:
Yes, so you mean we need kind of configurable swipe area when the panel is opened? I'm ok with this, can you add for left panel as well?

SpiderNeo1:
Exactly !
I named the property "swipeActiveAreaSide". If you want you can rename it otherwise.
Thank you for everything.

SpiderNeo1:
Vladimir,
It's good I added the left panel : https://github.com/framework7io/framework7/pull/2215/commits/93731e6c7c0cb55bfe6c0eadfe9cb122b8aa30d0

nolimits4web:
Yes, thanks, i will merge and test it later this week

SpiderNeo1:
Ok, thank you.

nolimits4web:
Merged, thanks 👍

PR #2230, opened 2018-02-28
Add capability to find old tab link with `data-route-tab-id` attribute…

Jule-:
… when using `tabs-routable`.

This commit allow to strongly find and update old `tab-link-active` in `tab-routable` scenarios thanks to `data-route-tab-id` attribute, even outside next siblings which is the actual way:
https://github.com/framework7io/framework7/blob/704695fed1c5e2a12dec3df4cfdf4c31d4c192d2/src/components/tabs/tabs.js#L149

In other words this allow to put tab links outside the toolbar if we want and rely on the good link activation of this component in order to maintain coherence of all links, wherever they are.


nolimits4web:
Thanks

Issue #2236, opened 2018-03-03
action sheet buttons background color not being applied


joseph-perez:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Framework7 Version: 2.0.10
* Platform and Target: Android, IOS

### What you did
Set the bg property for a action sheet button to "red" but the background color of the button does not change.

### Expected Behavior
The color set in the bg property for a button in an action sheet should change the background color of the button.

### Actual Behavior
The color set in the bg property for a button in an action sheet isn't changing the background color of the button. It's adding a class of bg-red instead of bg-color-red.

nolimits4web:
Fixed by #2237

PR #2237, opened 2018-03-03
fixing the background color property for action sheet buttons

joseph-perez:
The background color wasn't being applied correctly due to the class name being incorrect. I changed the class name to include -color- to fix the issue. I also noticed that the non popover option for the action sheet was using the color variable instead of the bg variable. I changed it to use the bg variable just like the popover version of the action sheet does.

I thought it might help so I included screenshots below of the code before and after the changes.

Before:
before
before-popover

After:
after
after-popover

Issue: https://github.com/framework7io/framework7/issues/2236


nolimits4web:
Thanks

PR #2246, opened 2018-03-05
add more accurate condition

zhennann:
In some special scene, when the `modal` closed, and  class `modal-out` also removed, then event `transitionEnd` may also be fired, which will invoke `modal.onOpened()`.


nolimits4web:
Thanks

PR #2253, opened 2018-03-06
feat: calendar-popover backdrop

zhennann:
When calendar show as popover, backdrop element can be configurable.


nolimits4web:
Good, thanks

PR #2259, opened 2018-03-07
Update README.md

TeijiW:
The **original link** returns error 404, so i changed for actual link for **"get started"**


nolimits4web:
Thanks

PR #2268, opened 2018-03-12
feat: support preloader within button

zhennann:
Shows preloader within button, such as:
```
<a class="button button-fill button-big">
        <div class="preloader">
          <span class="preloader-inner">
            <span class="preloader-inner-gap"></span>
            <span class="preloader-inner-left">
              <span class="preloader-inner-half-circle"></span>
            </span>
            <span class="preloader-inner-right">
              <span class="preloader-inner-half-circle"></span>
            </span>
          </span>
        </div>
      </a>
```

- Before fix:

- After fix:

### DIFF
--- a/src/components/button/button-md.less
+++ b/src/components/button/button-md.less
@@ -50,7 +50,7 @@
       }
     }
     i.icon + span,
-    span:not(.ripple-wave) + span,
+    span:not(.ripple-wave, .preloader-inner-gap, .preloader-inner-left, .preloader-inner-right) + span,
     span:not(.ripple-wave) + i.icon,
     i.icon + i.icon {
       .ltr({


nolimits4web:
Agree, but selector will be not so easy to overwrite, maybe let’s change it to just 

```
> span:not(.ripple-wave) + span
```
I mean to point the direct child span?

zhennann:
I tried:
```
.md .button .preloader .preloader-inner-left,
.md .button .preloader .preloader-inner-right{
  margin-left: 0 !important;
}
```
But not work!

nolimits4web:
I mean just adding `>` to the beginning of selector https://github.com/framework7io/framework7/commit/c6fb2c32999dcc61ccfac419a6966e8c7a6f8d85

zhennann:
Great!

PR #2277, opened 2018-03-15
Fix Accordion on iOS 9

SpiderNeo1:
Adaptation of the accordion of the V1 on the V2 to fix the bug when opening the accordion.


nolimits4web:
Thanks

PR #2300, opened 2018-03-24
Fix mobile Safari ignoring last non-alphabet after composition end

robinyip:
This fixed the buggy "input" event in iOS safari, which ignore the last non-alphabet (e.g. Chinese) character after composition end.

### DIFF
--- a/src/components/autocomplete/autocomplete-class.js
+++ b/src/components/autocomplete/autocomplete-class.js
@@ -232,7 +232,7 @@ class Autocomplete extends Framework7Class {
       }
       if (ac.params.openIn === 'dropdown' && ac.$inputEl) {
         ac.$inputEl.on('focus', onInputFocus);
-        ac.$inputEl.on('input', onInputChange);
+        ac.$inputEl.on('change keyup compositionend', onInputChange);
         if (app.device.android) {
           $('html').on('click', onHtmlClick);
         } else {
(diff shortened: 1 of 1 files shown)


robinyip:
rather than change the event listener, add a timeout 0 for the "input" event also work and this should be better

robinyip:
How should I fix below travis error:

`Unexpected function expression  prefer-arrow-callback`

nolimits4web:
Don't really like an idea of using setTimeout, also such dynamic event listener function with setTimeout can't be removed. Can you give me some steps on how to replicate the issue? I will try to look for a better fix

robinyip:
Please tried the following in mobile Safari or cordova in iOS:
1. create an <input> field.
2. Bind the "input" event of that input field to log the field value in console or write on screen.
3. When you use IME input method (e.g. Chinese) to composite a Chinese character, you can see your original composited Chinese character is logged and immediately follow by a value logged with the last Chinese character missing.

The above cause the autocomplete dropdown ignoring the last Chinese character. It can be either fixed by not using the "input" event as following:

ac.$inputEl.on('change keyup compositionend', onInputChange);

Hope the above explained clearly, let me know if anything is unclear, thanks.

nolimits4web:
I see now. Maybe we can just make `inputEvent` as a parameter which is by default is “input” and can be changed for such cases to “change keyup compositionend”? I’m afraid we can break some edge cases if we change it to this in core

robinyip:
Agree with you, it can be an option if anyone facing the same problem

PR #2308, opened 2018-03-29
Support non string className (SVGAnimatedString)

safrazik:
The following error occurs when swipeout item is opened and an outside SVG element is cliked
```
Uncaught TypeError: $targetEl[0].className.indexOf is not a function
```
It appears that $targetEl[0].className is an instance of [SVGAnimatedString](https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedString) when clicked on an SVG element. 

as a workaround I do the following, which is not a standard practice in a long run:

```javascript
   SVGAnimatedString.prototype.indexOf = function(){
     return this.baseVal.indexOf.apply(this.baseVal, arguments)
   }
```

getting the attribute directly without the `className` property supports both html elements and svg elements.


nolimits4web:
Looks good, thanks!

PR #2317, opened 2018-03-30
Add ptr support for pages with subnavbar/toolbar

safrazik:
On material theme, when you have a subnavbar or toolbar/tabbar along with pull to refresh, the preloader hides behind the subnavbar or toolbar.

This pull request fixes that issue.


safrazik:
In iOS this issue is not there because it uses relative positioning.

nolimits4web:
Thanks, merged!

Issue #2324, opened 2018-04-03
List Index scroll upward display problem


robinyip:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Framework7 Version: 2.2.0
* Platform and Target: Any

### What you did
Click on an List Index item which the list view element is above the current scrolling position so the list have to scroll upward

### Expected Behavior
The `list-group-title` of the clicked List Index value should be position at the top and follow by the `list-group` content

### Actual Behavior
Only the `list-group-title` is positioned at top but the group content is not visible

nolimits4web:
Fixed by #2325

PR #2325, opened 2018-04-03
Fix scrollTop value when scroll upward

robinyip:
Fix the scrollTop value when the clicked index required to scroll upward, issue #2324


nolimits4web:
Looks good, but let's cache these dom calcs before if statement, to prevent them from calculated twice, like
```
const parentTop = $scrollToEl.parent().offset().top;
const paddingTop = parseInt($pageContentEl.css('padding-top'), 10);
...
if (parentTop <= paddingTop) {
...
}
```

nolimits4web:
Man, and i meant to use these cached vars inside too :) E.g.
```
$pageContentEl.scrollTop((parentTop + scrollTop) -paddingTop);
```
Using these cached vars instead of calculating offsets will improve performance

nolimits4web:
Great, thanks!

PR #2370, opened 2018-05-07
Navigation and old pages reference fix

llaszkie:
framework7io/framework7-vue#412

Including:
* a proper selection of available pages 
* cut off an old long chain of references to `fromPage`. The cut fixes the memory leak problem.

### DIFF
--- a/src/modules/router/clear-history.js
+++ b/src/modules/router/clear-history.js
@@ -9,7 +9,7 @@ function clearHistory() {
   const $currentPageEl = $(router.currentPageEl);
 
   const $pagesToRemove = router.$el
-    .children('.page:not(.stacked)')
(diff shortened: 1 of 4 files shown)


nolimits4web:
I think i am good with pageFrom deep reference removal, but have a question about `'.page:not(.stacked), .page-current:not(.stacked), .page-previous:not(.stacked), .page-next:not(.stacked)'`  selector. What is the advantage? `.page-current:not(.stacked)` will anyway match to `.page:not(.stacked)`, because the page can't exist with just a `.page-current` class, it is always like `.page.page-current` or `.page.page-next`

nolimits4web:
Merged, thanks!

Issue #2379, opened 2018-05-10
app.panel.disableSwipe throwing error


chachew:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Framework7 Version: 2.2.5
* Platform and Target: Cordova

### What you did
In previous parts of my code i am enabling the right panel swipe option. `app.panel.enableSwipe('right');`
Later in my code after i am done with the right panel i want to disable the right panel swipe action. `app.panel.disableSwipe('right');`

### Expected Behavior
I expect the system to disable the swipe action without issue.

### Actual Behavior
A js error is thrown in the console and it then locks up the rest of my application.

`Uncaught TypeError: panels.forEach is not a function
    at Object.disableSwipe (framework7.js:14811)
    at Framework7.<anonymous> (my-app.js:273)
    at framework7.js:3047
    at Array.forEach (<anonymous>)
    at framework7.js:3046
    at Array.forEach (<anonymous>)
    at Framework7.emit (framework7.js:3040)
    at framework7.js:3053
    at Array.forEach (<anonymous>)
    at View.emit (framework7.js:3052)`

nolimits4web:
Fixed by #2380

PR #2380, opened 2018-05-10
Issue #2379

chachew:
Fix js error when calling app.panel.disableSwipe('left'); OR app.panel.disableSwipe('right');

This fixes the javascript errors that are thrown when trying to call app.panel.disableSwipe('left') OR app.panel.disableSwipe('right');


nolimits4web:
Thanks, but this need to be done for `src` files only:

https://github.com/framework7io/framework7#contributing

> All changes should be commited to src/ files.

chachew:
Sorry, fixed that.

nolimits4web:
Changes to `dist/` file should be removed

nolimits4web:
Thanks 👍

PR #2393, opened 2018-05-16
Adding discord channel for live help support

rlam3:
Adding discord channel for live help support


nolimits4web:
Closing as there is already forum which is a great place for sharing and helping

PR #2414, opened 2018-05-31
Add Jalali calendar to F7, Add Min & Max days to Date Range Picker

sepehrr:
Hi
Iranian use Jalali Calendar instead of Gregorian and it is essential for them to provide jalali calendar in their application. for that reason some of other frameworks like Material UI have that component (https://material-ui-pickers.firebaseapp.com/localization/persian)
So I used IDate(https://www.npmjs.com/package/idate) and make jalali calendar available for F7 with minimum code change in F7's calendar component.
Just put `calendarType: 'jalali'` as a parameter when creating new calendar is enough.

Another thing that I added to calendar, is minimum and maximum days that can be selected in range picker. By default there is no restriction but by providing value for `minimumDaysInRange` and `maximumDaysInRange` we can achieve this functionality.

I added Calendar Jalali and max/min days to kitchen sink too, so you can check them. (I can confirm that Jalali is working flawlessly)


nolimits4web:
A lot of work, thanks, but would be good to discuss it first. Few things:

1) First of all, all commits must be done to files in `/src/` folder only
2) Don't really like the idea of having one more 3rd party dependancy for this. Would it be too hard to implement it from scratch or reuse some code of the libs, without having to include full lib?

sepehrr:
1. All right, I will do that. Changes in other places(except in Kitchen sink that I added demo) was done when I built the project and they can be undone.
2. the good thing about IDate class is all of its functions are same as Date class so changes would be minimal and anything you do with Date class, you can do with IDate. Maybe we can use IDate or JDate (https://github.com/tahajahangir/jdate) dist file.

sepehrr:
@nolimits4web I close this pr and after v3 release, I will make another pr based on V3

Issue #2423, opened 2018-06-08
[V2] Autocomplete problem when dropdown item with double quote


robinyip:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* Framework7 Version: 2.2.5
* Platform and Target: All

### What you did
Dropdown item with double quote

### Expected Behavior
Should work properly

### Actual Behavior
The value parameter in “change” event will be "undefined" when that item is selected

nolimits4web:
Fixed by #2424

PR #2424, opened 2018-06-08
Fixed problem when dropdown item contains double quote

robinyip:
Fixed the problem with data-value attribute when the dropdown item contains double quote
#2423

### DIFF
--- a/src/core/components/autocomplete/autocomplete-class.js
+++ b/src/core/components/autocomplete/autocomplete-class.js
@@ -436,12 +436,12 @@ class Autocomplete extends Framework7Class {
     const ac = this;
     if (ac.params.renderItem) return ac.params.renderItem.call(ac, item, index);
     let itemHtml;
-    item.value = item.value.replace(/"/g, '&quot;');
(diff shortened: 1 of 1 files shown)


nolimits4web:
Looks good to me with replacement, but let's just instead of modifying current item's `item.value` do the new var like `const itemValue = ...` and use it further in code?

robinyip:
Updated with new variable. BTW, may I know what's the advantage of assigning to a new variable?

nolimits4web:
Thanks! Not really advantage, but modifying passed arguments considered harmful to avoid side effects. For example in situations like this:
```js
var item = {
  value: 'fo"o',
};

function renderItem(item) {
  item.value.replace = item.value.replace(/"/g, '&quot;');
  ...
}
function doSomethingWithItem(item) {
  // this will be false after render item
  if (item.value.indexOf('"') >= 0) {
    ...
  }
}

renderItem(item);
// this may not work as expected because we modified item value
doSomethingWithItem(item); 
```

PR #2431, opened 2018-06-12
[v3] [React] Update list-item-content.js - Prevent warning to fire out

AliMamed:
Prevent warning to fire out for ListItemContent. 
`setState` shouldn't appear in `render()` but may appear in `componentWillMount` and `componentWillUpdate`


nolimits4web:
Thanks! I have merged it manually as this fix must be done in Phenome component source code instead

PR #2435, opened 2018-06-14
Support for SmartSelect in Link

AliMamed:
SmartSelect was supported within `<a>` tag in v2, so I think it would be nice to have it in v3 also


nolimits4web:
It wasn't support in v2 under `<Link>` component, but i don't mind, let's have it here :)

PR #2476, opened 2018-07-02
Use default behavior of opening a external link

tylerlong:
Use default behavior of opening a external link

Why not `window.open`?  Because some times browser blocks it.  

If it is just a normal external link, why not use the default behavior?


nolimits4web:
Thanks

PR #2477, opened 2018-07-02
Use default behavior to open external link

tylerlong:
Use default behavior to open external link

Why not `window.open`? Because some browsers block it

If we know it is just a simple normal external link, why not just open it using the default behavior?


nolimits4web:
Merged, thanks

tylerlong:
Do you have plan to release a new version of V2?

PR #2521, opened 2018-07-26
specify more precisely which popup to close

AliMamed:
If smartSelect is opened in a popup from other popup (sourcePopup), then clicking on close button in smartSelect popup header will cause the sourcePopup to be closed, while smartSelect popup remains opened.

I propose to more precisely set which popup to close in smartSelect renderPopup function


nolimits4web:
Thanks! 👍

PR #2523, opened 2018-07-27
popover.open() should return popover

AliMamed:
I think it was a typo. If modal is not returned here closeOnSelect f.e. doesnt work as expected


nolimits4web:
Merged, thanks!

PR #2554, opened 2018-08-04
 better swipeback on rtl theme 

sepehrr:
I faced a problem that swipe-back on rtl theme triggered very hard! you should swipe from very very right edge and sometimes it is triggered and sometime(most of the time) no luck. 
I checked the `swipe-back.js` and with changing this line problem seems to be solved and swipe-back on rtl triggered smoothly


nolimits4web:
Thanks! 👍

PR #2555, opened 2018-08-05
Update toolbar.js

zhennann:
In some specific scenario, the count of tabLinks may be changed from some to zero, which will need remove the highlight element.


nolimits4web:
Thanks! 👍

PR #2560, opened 2018-08-07
Update clicks.js

zhennann:
If customize the link behaviors, the system click-handle always be performed. So maybe add class `no-auto` to provide a chance to disable the system click-handle.

### DIFF
--- a/src/core/modules/clicks/clicks.js
+++ b/src/core/modules/clicks/clicks.js
@@ -12,6 +12,9 @@ function initClicks(app) {
     const url = isLink && clickedLink.attr('href');
     const isTabLink = isLink && clickedLink.hasClass('tab-link') && (clickedLink.attr('data-tab') || (url && url.indexOf('#') === 0));
 
+    // Check if no-auto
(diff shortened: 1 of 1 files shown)


nolimits4web:
But `class="external"` will do the same, no?

bencompton:
I was actually coming in to do a similar PR, except I was going to add an overall app parameter (need to disable automatically invoking the router for links to make Framework7 Redux work). Now that you mention it, @nolimits4web , I guess this can also be accomplished at the app level by setting `externalLinks` to `#`.

nolimits4web:
@bencompton yes, it match to CSS selector specified, so it can be just `externalLinks: 'a'` to ignore all links or `externalLinks: 'a[href="#"]'` to ignore links with such href, etc.

bencompton:
@nolimits4web - thanks, the `externalLinks` setting does work perfectly. Disabling the automatic routing behavior is not going to be used by many people, so I guess I can just make sure the Framework7 Redux docs document this appropriately.

zhennann:
`class="external"` works nice!  thanks!

PR #2582, opened 2018-08-16
add className param to smart-select

Fire7:
In some cases when you need to customize layout of smart-select `render` methods is overhead, it is just some styles changes, so className option is here for help.


Fire7:
@nolimits4web done

nolimits4web:
Thanks ! 👍

PR #2584, opened 2018-08-16
Routable panels

bencompton:
This PR adds the capability of creating, opening, and closing panels via routes.

I discovered that the reason routable panels probably weren't already implemented is because creating dynamic panels is currently not possible. F7 is hard-coded all over the place to expect only a left panel or a right panel. In order to support routable panels, I did add the ability to create dynamic panels (from a content string, component, etc.), but rather than risk creating a lot of bugs, I decided to be more conservative and simply throw an error when adding a dynamic panel to a side that already contains a panel.

So, for all practical purposes, people will need to either use dynamic panels or non-dynamic panels, but not both. I thought this was a reasonable compromise, even though it would certainly be nice to support both like modals do. I added examples to the kitchen sink, but will need to revert them before this PR is merged because the kitchen sink already has non-dynamic panels (I temporarily removed them to make the routable panel example work). Once this PR is merged, I'll create a PR to update the docs.

This routable panel functionality leverages the existing routable modal code. I also fixed a [bug in that code](https://github.com/framework7io/framework7/commit/7d131e247d5abd80a51fde5b06bf10c7668398cd) where modals / panels could only close automatically when navigating back, but not navigating forward to another route.


nolimits4web:
Hey Ben! This is awesome. I thought it would be harder to implement, nice catch with reusing modal loading logic for panels. There are couple of micro issues i found. Let's fix them and i'll merge it. After i'll try to see what can be done to allow to use both dynamic and non-dynamic panels like with modals. so we should be able to use both static and routable panels

bencompton:
Hey, thanks for looking over these changes! I went ahead and made the changes you suggested.

nolimits4web:
Awesome, thanks!

nolimits4web:
I also reverted Kitchen sink to original panels for now

PR #2606, opened 2018-08-25
f7-list - add form submit event

natiz:
This would allow:

```
<f7-list form @submit="myOnSubmitMethod">...</f7-list>
```

While currently what I do is:
```
<f7-list form ref="form">...</f7-list>
...
  mounted() {
    this.$nextTick(() => {
      this.$refs.form.$el.addEventListener('submit', this.myOnSubmitMethod);
    });
  },
```


nolimits4web:
Thanks!

PR #2608, opened 2018-08-25
add `backdrop` option for `calendar`

sepehrr:
Hi
Right now `backdrop` is available only when calendar's type is popover, but when it is`customModal` or `sheet` it may be needed too. (I myself needed this for `customModal` badly!)
I added `closeByBackdropClick` too.

### DIFF
--- a/src/core/components/calendar/calendar-class.js
+++ b/src/core/components/calendar/calendar-class.js
@@ -1356,6 +1356,7 @@ class Calendar extends Framework7Class {
       scrollToEl: calendar.params.scrollToInput ? $inputEl : undefined,
       content: modalContent,
       backdrop: calendar.params.backdrop,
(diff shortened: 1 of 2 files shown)


nolimits4web:
Ok, but by default it should be the same as before to not break anyones behavior, 
```
modalType === 'popover' && app.params.popover.backdrop !== false
```

sepehrr:
I made some changes.
By default `backdrop` is null so:
1. if `backdrop` is not set by user then everything is like before. (only backdrop for `popover` when `app.params.popover.backdrop` is `true`)
2. if it is set `true`, then backdrop would be displayed(and for `popover`, the `app.params.popover.backdrop` is ignored).
3. if it is set `false`, there would be no backdrop(and for `popover`, `app.params.popover.backdrop` is ignored).

nolimits4web:
Great, thanks! 👍

PR #2613, opened 2018-08-27
add backdrop and animate props to popup

jascolley:
Needed access to these two props on the popup in order to disable them in our app.


nolimits4web:
Thanks! 👍

PR #2638, opened 2018-09-05
optimizing toolbar.js “no-toolbar”

chrysanthemum:
"toolbar div" and "no-toolbar property" exist together pageBeforeIn,  all the other "toolbars div" should hide,  cover  `page.$el.parents('.views').children('.tabbar, .tabbar-labels')`.

if not,    .views > .tabbar, .tabbar-labels can't hide().

please give your comment. tks.


nolimits4web:
Makes sense, thanks!

PR #2666, opened 2018-09-14
Add missing interface props for RouteOptions for router.back

dlee:
Added `ignoreCache` and `force` to RouteOptions interface.


nolimits4web:
Thanks! 👍

PR #2668, opened 2018-09-14
Fix parameter typing for Router.navigate

dlee:
Right now, you can't use `Router.navigate()` with a named route because its parameter is declared as a `Route` with all options required. I added a `NavigateParameters` interface that limits that to a `name`, `query?`, and `params?`.


nolimits4web:
Great, thanks! 👍

PR #2669, opened 2018-09-15
Update Less to 3.x

### LINKED ISSUE #2285: [v4] [less]  inline javaScript is deprecated / update to less 3.x
This is a (multiple allowed):
* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

Starting with less 3.0.0 Inline JavaScript is Deprecated in the flavor of plugins. ([Docs](http://lesscss.org/usage/#less-options-enable-inline-javascript-deprecated-))

It is currently being used for RTL variant. Example [list.less:196](https://github.com/framework7io/framework7/blob/master/src/components/list/list.less#L196)

BTW this is just an informational issue. Please close if there are no any plans using less 3 :D

fjorgemota:
Hey!

First, thank you for the project. It seems really great. =)

This PR closes #2285, by migrating all inline JS code of the less files to a Less plugin, that registers some functions dedicated to framework7 on Less namespace and executes the same exact code of the original inline JS code, so maintaining the behavior.

I've tested the PR here by running the kitchen sink and seems pretty ok..So I think everything is ok. I'm not sure if this PR could be merged even on the 3.x series..but anyway here it is for the future of the project. =)

Any questions or problems please feel free to ask.

Thanks. =)


nolimits4web:
Awesome, thanks, merged! 👍

nolimits4web:
Yeah, noticed it too, update package-lock and now it is ok

fjorgemota:
Nice! Thanks! :)

PR #2697, opened 2018-09-21
dominantBaseline should not have boolean value

AliMamed:
dominantBaseline should not have boolean value.
In case it does we get react warning

In this small change I propose to have `null` instead of `false` for semiCircle


nolimits4web:
Great, thanks! 👍

PR #2698, opened 2018-09-22
Typescript fixes for compilation errors

ariofrio:
It doesn't seem like framework7 and framework7-react were compiling under TypeScript 3.0.1. Looking at the problems, it doesn't seem to me that they were compiling under any version of TypeScript, but I haven't confirmed that.

These changes aim to fix that. I have a project that is compiling successfully with these changes plus https://github.com/phenomejs/phenome/pull/40.

Ideally, we'd at least add a smoke test that just imports framework7 and framework7-react from a TypeScript file and makes sure that it compiles.

Related: https://github.com/phenomejs/phenome/issues/11

### DIFF
--- a/packages/core/components/actions/actions.d.ts
+++ b/packages/core/components/actions/actions.d.ts
@@ -86,7 +86,7 @@ export namespace Actions {
     renderPopover?: () => string
     /** Object with event handlers */
     on?: {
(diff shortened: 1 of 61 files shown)


nolimits4web:
Awesome, thanks! 👍

PR #2699, opened 2018-09-22
Compatibility with yarn workspaces

ariofrio:
In my project, I use yarn workspaces (and git submodules) to maintain my own branch of some dependencies. For example, there are [some fixes](https://github.com/phenomejs/phenome/issues/11) to framework7's TypeScript bindings that I needed to make for it to compile. The directory structure is like so:

```
app/
  packages/
    framework7/
      build/
        core/
        react/
        vue/
    app/
```

Yarn Workspaces ensures that when `app` requires e.g. `framework7-react`, it's using the one in the packages/ folder. This is an improvement over `npm link` because it doesn't break when `npm install` is used.

To make this work, I made the following changes:

- change root package name so it doesn't conflict with core package
- add 0.0.0 version so (with some effort) this can used as a yarn workspace, works around https://github.com/yarnpkg/yarn/issues/5936

Since the changes are minimal and might make things more clear anyway, you might consider merging them upstream.

<details>
<summary>For reference, here's the rest of my setup:</summary>

**scripts/framework7-dev-packages.js:**

```.js
// There are two ways to build Framework7. The development build (build/) is
// ignored by Git, comes with a watch mode, but doesn't include package.json and
// root typings. The production build (packages/) shows in `git diff`, has no
// watch mode, but includes package.json and root typings. The production build
// is supposed to only be updated when a new Framework7 verision is released.
//
// This script makes the development build usable as yarn workspaces.

const fs = require('fs-extra');
const path = require('path');

const srcDir = 'packages/framework7/packages';
const dstDir = 'packages/framework7/build';

if (fs.existsSync(dstDir)) {
  fs.removeSync(dstDir);
}
fs.mkdirSync(dstDir);

for (const pkg of ['core', 'react', 'vue']) {
  const srcPkgDir = path.join(srcDir, pkg);
  const dstPkgDir = path.join(dstDir, pkg);
  const srcFiles = [path.join(srcPkgDir, 'package.json')];

  if (!fs.existsSync(dstPkgDir)) {
    fs.mkdirSync(dstPkgDir);
  }
  for (const srcFile of srcFiles) {
    fs.copyFileSync(srcFile, path.join(dstPkgDir, path.basename(srcFile)));
  }
}

```

**package.json:**

```.json
{
  "name": "root",
  "private": true,
  "workspaces": {
    "packages": [
      "packages/*",
      "packages/framework7/build/*"
    ]
  },
  "scripts": {
    "framework7-dev-packages": "node scripts/framework7-dev-packages.js",
    "build-deps": "yarn framework7-dev-packages && yarn workspace framework7-source build:dev && yarn"
  },
  "devDependencies": {
    "fs-extra": "^7.0.0"
  }
}
```

</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
-  "name": "framework7",
+  "name": "framework7-source",
+  "version": "0.0.0",
   "scripts": {
     "build:dev": "cross-env-shell NODE_ENV=development \"gulp build-core && gulp phenome && gulp build-vue && gulp build-react\"",
     "build:prod": "cross-env-shell NODE_ENV=production \"gulp build-core && gulp phenome && gulp build-vue && gulp build-react\"",


nolimits4web:
Thanks! 👍

PR #2704, opened 2018-09-22
cross-env-shell for more scripts

JasonKleban:
I think this is necessary or else these scripts, run alone, do not output to the packages folder.


nolimits4web:
Right, thanks!

PR #2722, opened 2018-09-26
Fix Accordion on iOS 9

SpiderNeo1:
Adaptation of the accordion of the V2 on the V3 to fix the bug when opening the accordion (as I did on v2 https://github.com/framework7io/framework7/pull/2277).

Thank you.


nolimits4web:
I removed it intentionally as `.nextFrame` now must do the thing. If it is still the issue for iOS 9 only then we should target it for iOS 9 only, as requesting `.clientLeft` requires extra resources

SpiderNeo1:
Ok, I found the solution. I just tested and it works well on iOS9 and above.

nolimits4web:
Fixed by #2731

PR #2725, opened 2018-09-26
add core/react typings to build

JasonKleban:
I think these are good enough to start.  The current release has syntax errors in it, but the actual source is right, it just wasn't included in the build.  This should avoid that in the future.

(I know vue isn't done, but I think that's a different animal.  Anybody who cares about typesafe vue can advise on the appropriate code to generate there.  I tried vue briefly and switched back to react because it had better tooling support and was thus easier to make typesafe, so I don't know how to do that one.)


nolimits4web:
@JasonKleban there are no such `build-core/react/vue-typings` gulp tasks, and such build-typings commands are already included into `build-core/react/vue` tasks https://github.com/framework7io/framework7/blob/master/scripts/gulpfile.js#L42-L44

JasonKleban:
Hmm, ok.  I was able to run this successfully many times - not sure how - but I see what you're saying about it.  There's just something still wrong for me when I try to build:prod that isn't including the types, I think, but the problem must be somewhere else.  Sorry.

PR #2727, opened 2018-09-27
added El to callbackData properties

itsimon21:
Swipe Back Events
http://framework7.io/docs/view.html#router-instance-events


nolimits4web:
Thanks! 👍

PR #2731, opened 2018-09-27
Fix bug Accordion on iOS 9 (using nextFrame function)

SpiderNeo1:
Fix the bug when opening the accordion on iOS 9 using the function `nextFrame`.

Thank you Vladimir.


nolimits4web:
Looks good, thanks!

PR #2739, opened 2018-09-30
Do not clone deep options in forward

blikblum:
This is a followup to #2713.

In the process of cloning options in forward function, the the route instance is replaced by a copy, so the route instance stored in `router.currentRoute` is different from the one passed to route async handler. 

The proposed change is safe since `options` variable is not mutated at all.

It is also more performant


nolimits4web:
Looks good, thanks!

PR #2746, opened 2018-10-02
Prevent smartSelect to be rendered twice

AliMamed:
This change is similar to what you did in https://github.com/framework7io/framework7/commit/1973c08ed99db3848ee943740bb2e5e656a15a17

Currently smartSelect attached to link renders two popups on link click.


nolimits4web:
Good, thanks!

Issue #2841, opened 2018-11-19
Allow programmatic clearing of Dom cache.


mlitwin:
The use case here is when user switches language in a multilingual application, it would be convenient to be able to just clear the cache. That way your can use the existing router.refreshPage() to update the current page, and then a new router.clearDomCache() to prepare pages in history to be updated to correct language when navigated to.

Looking at the implementation of refreshPage(), the first thing it does is clear the DOM cache, so I think it is just factoring out that implementation from refreshPage() into a new exported function. I'll put up a PR for your consideration.

PR #2842, opened 2018-11-19
Implement clearDomCache method in the Router

mlitwin:
Factor out the Dom cache clearing from clearPreviousHistory(), and export it as a Router method.

See https://github.com/framework7io/framework7/issues/2841

This facilitates a use case like chaning language preference, where you can refreshPage() to reload the current page in the current language, and then call clearDomCache() to ensure that pages in the history will be reloaded in the current language if they are revisited.

Always follow the [contribution guidelines](https://github.com/framework7io/framework7/blob/master/.github/CONTRIBUTING.md) when submitting a pull request.

### DIFF
--- a/src/core/modules/router/clear-previous-history.js
+++ b/src/core/modules/router/clear-previous-history.js
@@ -1,6 +1,6 @@
 import $ from 'dom7';
 
-function clearDomCache() {
+function clearStackedPages() {
   const router = this;
   const app = router.app;
   const separateNavbar = router.separateNavbar;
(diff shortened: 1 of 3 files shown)


nolimits4web:
Ok, got the point. But let's:

* rename it to `clearStackedPages`
* It also needs to remove page with `.stacked` class in both `clearStackedPages` and `clearPreviousHistory`, e.g. change `.children('.page:not(.stacked)')` to just `.children('.page')`

mlitwin:
Ah just noticed you asked for "clearPreviousPages" ...

nolimits4web:
Great, thanks!

PR #2845, opened 2018-11-20
bind full event object instead of prevent() reference only

kozmanbalint:
Bind full event object to beforeOpen and beforeClose events in accordion-item and list-item phenome component to make prevent() actually useful.


nolimits4web:
Thanks, i added `prevent()` as second argument

PR #2870, opened 2018-12-05
Browser field in package.json

Fire7:
Webpack has `mainFields` option to resolve modules.
https://webpack.js.org/configuration/resolve/#resolve-mainfields

By default it is `['browser', 'module', 'main']`,  so `module` has more priority than `main`. 
Obviously, for web and mobile devices es5 is more preferably (iPhone 5 @ ios 10, Android < 5), so in order not to change the configuration of webpack, I suggest to add `browser` field same as `main`.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -3,6 +3,7 @@
   "version": "3.5.2",
   "description": "Full featured mobile HTML framework for building iOS & Android apps",
   "main": "js/framework7.js",
+  "browser": "js/framework7.js",
   "typings": "framework7.d.ts",
   "jsnext:main": "framework7.esm.js",
   "module": "framework7.esm.js",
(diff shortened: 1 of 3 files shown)


mmis1000:
I believe when using webpack, original esm/cjs version should be used as source, and use babel..etc to convert it for low version bowser.

Use packed version in webpack will break the source map completely, the debug no longer show the origin of source, it will instead show the position in packed version, which is no one that expect it to be.

nolimits4web:
Sorry, it is not an option. Webpack is module loader/bundler so in order to utilize tree shaking and ability to import only components you need like `import {Swiper} from 'framework7'` it must import `.esm` module. So if you don't need any module-related features you may explicitly import browser version like `import Framework7 from 'framework7/js/framework7.js';`. But it is not recommended and it is better to configure babel to transpile Framework7 module too

PR #2884, opened 2018-12-12
[Kitchen Sink] Icons with a long name cause overflow problems

DAnn2012:
The long name of some icons causes overflow problems in the Demo Page of Icons on devices that have a small screen


nolimits4web:
Thanks

Issue #2885, opened 2018-12-12
Relax React and Vue global TypeScript definitions for Framework7Extensions


jmca:
* Framework7 version: 3.6.2
* React.js version: 16.6.3
* Platform and Target: TypeScript related

### Describe the bug
The TypeScript definition of `Framework7Extensions` is written as:
```
declare module 'react' {
    interface Component extends Framework7Extensions {}
}
```

(located in files `framework7-react.d.ts` and `framework7-react.esm.d.ts`)

Since Framework7 is extending React's Component type globally, TS is throwing errors like:
```
  Type 'ReactElement<{}>' is not assignable to type 'ReactElement<any>'.
    Types of property 'type' are incompatible.
      Type 'string | ComponentClass<{}, any> | StatelessComponent<{}>' is not assignable to type 'string | ComponentClass<any, any> | FunctionComponent<any>'.
        Type 'ComponentClass<{}, any>' is not assignable to type 'string | ComponentClass<any, any> | FunctionComponent<any>'.
          Type 'ComponentClass<{}, any>' is not assignable to type 'ComponentClass<any, any>'.
            Type 'Component<{}, any, any>' is not assignable to type 'Component<any, any, any>'.
              Property '$f7' is missing in type 'Component<{}, any, any>'.
```

The understandable part of the error:
```
Property '$f7' is missing in type 'Component<{}, any, any>'.
```

I'm proposing a more relaxed Component extension using the following:
```
declare module 'react' {
    interface Component extends Partial<Framework7Extensions> {}
}
```

This will play nicer with existing React/React-3rd-party definitions since the additional extension properties would be optional and not required.

This issue also applies to Vue's TS definitions.

nolimits4web:
So is it changing line
```
interface Component extends Framework7Extensions {}
```
to
```
interface Component extends Partial<Framework7Extensions> {}
```
?

Will/can it affect something?

jmca:
I would consider this a non-breaking change, since it’s only making `Framework7Extensions` properties optional instead of required. I’m open to other thoughts though.

Here's some verbiage on `Partial`:

https://netbasal.com/getting-to-know-the-partial-type-in-typescript-ecfcfbc87cb6

(Somewhere on the page)
https://www.typescriptlang.org/docs/handbook/advanced-types.html

PR #2899, opened 2018-12-19
#2885 Make TS Component extension properties optional via Partial

jmca:
Fixes issue #2885

### DIFF
--- a/packages/react/framework7-react.d.ts
+++ b/packages/react/framework7-react.d.ts
@@ -117,7 +117,7 @@ export interface Framework7Extensions {
 }
 
 declare module 'react' {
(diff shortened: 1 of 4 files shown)


nolimits4web:
Thanks, but changes in `packages` folder can't be accepted as it modifies exisiting releast. Can you please just do same changes in:

https://github.com/framework7io/framework7/blob/master/scripts/build-react-typings.js#L17
and
https://github.com/framework7io/framework7/blob/master/scripts/build-vue-typings.js#L17

jmca:
@nolimits4web Ah, I see what you're saying, I'll change that now.

nolimits4web:
Merged, thanks!

PR #2921, opened 2018-12-29
[KS] Long $route.params causes overflow problems

DAnn2012:
The long strings ($route.params) causes overflow problems in the Page Loaders examples (Template7 Page and Component Page) on devices that have a small screen


nolimits4web:
Thanks!

PR #2933, opened 2019-01-08
fix Gulp error: The following tasks did not complete: Did you forget …

yelban:
update `./scripts/gulpfile.js` to prevent Gulp 4.0.0 error when stopping gulp server.


nolimits4web:
Thanks

PR #2936, opened 2019-01-08
Fix vue ks

scriptPilot:
Fixes three points in the Vue ks - see details in the commits :-)

### DIFF
--- a/kitchen-sink/vue/css/app.css
+++ b/kitchen-sink/vue/css/app.css
@@ -79,10 +79,6 @@
   background: url(../img/f7-icon.png) no-repeat center;
   background-size: cover;
 }
-.icon-vi {
-  background: url(../img/vi-icon.png) no-repeat center;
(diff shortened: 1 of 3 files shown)


scriptPilot:
@nolimits4web - What about the other ESLint fixes?

nolimits4web:
@scriptPilot Kitchen Sink is in eslintignore so no linting required there

scriptPilot:
@nolimits4web  - But what is bad on it, fixing these small points, to be in line with the Style Guide in the kitchen sink as well?

nolimits4web:
@scriptPilot because they won't fix much, and the more important, not everyone understand what is this `[theme] = document.location.search.split('theme=')[1].split('&');`. Syntax there must be as much as simple as possible. Not everyone aware of ES-next stuff

scriptPilot:
@nolimits4web - Ok, get the point ...

PR #2946, opened 2019-01-16
Fix compile error in Typescript Strict Mode

Hizoul:
If typescript is in strict mode everything needs to be typed explicitly. In Line 16 of this file there was one occurrence where that was not the case which leads to the following error:
```
ERROR in [at-loader] ./node_modules/framework7/modules/router/router.d.ts:16:37
    TS7019: Rest parameter 'args' implicitly has an 'any[]' type.
```


nolimits4web:
Thanks!

PR #2984, opened 2019-02-06
Remove webkit prefix for animation properties and events

blikblum:
Removes webkit prefixes for animation properties and events. It simplifies code, reducing number of registered events

All supported browsers in upcoming v4 supports unprefixed versions

Keeps WebKitCSSMatrix since the standard DOMMatrix is not widely supported. Also kept the support checks

### DIFF
--- a/src/core/components/tooltip/tooltip-class.js
+++ b/src/core/components/tooltip/tooltip-class.js
@@ -79,7 +79,7 @@ class Tooltip extends Framework7Class {
     }
 
     tooltip.attachEvents = function attachEvents() {
-      $el.on('transitionend webkitTransitionEnd', handleTransitionEnd);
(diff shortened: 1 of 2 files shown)


nolimits4web:
Thanks, but can you please revert changes in Swiper? It has separate repo where it can't be removed for now

blikblum:
Done.

BTW: the unprefixed support for animation/transition events landed on ios safari v9, which was released in 2015. iOS v9 update is available up to iPhone4S so its support still covers ageing devices

Let me know if/when drop support for iOS < v9, so i can patch dom7 and swiper

nolimits4web:
@blikblum Swiper (and Dom7) are still indented to be used on some older browsers including older androids, so let's keep them there

PR #3006, opened 2019-02-12
Closes #3005 by using the correct relative path while generating .d.ts files

### LINKED ISSUE #3005: framework7.d.ts in the js/ folder has wrong relative path
* Framework7 version: 4.0.1
* Platform and Target: Linux, any. (the error is caused directly on Typescript compilation)

### Describe the bug
There are 4 "main" `.d.ts` file in the final build sent to `npm`, written [here](https://github.com/framework7io/framework7/blob/master/scripts/build-core-typings.js#L55-L58).

The problem is that, while the `.d.ts` saved on the root folder of the package are correct, the files in the `js/` folder aren't, because they continue to refer to the components in the same way as the `.d.ts` files saved in the root folder of the package, doing imports like this:

```js
import {Accordion as AccordionNamespace} from './components/accordion/accordion';
```

While obviously the `components` folder isn't inside the `js/` folder so the typescript compilation fails. To finalize, the `typings` field in `package.json` has the value `js/framework7.d.ts`, which causes any typescript project using Framework7 4 to fail.

### To Reproduce
1. Create a project and install `framework7` and `typescript`;
2. Create a typescript file importing framework7;
3. Try to compile the file, it will fail because of the path problem reported above;

### Expected behavior
Typescript files importing Framework7 should work fine.

### Actual Behavior
Typescript emit a bunch of errors about files in `./components` folder (and other folders too) not found.

fjorgemota:
See #3005

### DIFF
--- a/scripts/build-core-typings.js
+++ b/scripts/build-core-typings.js
@@ -16,6 +16,32 @@ function capitalize(name) {
   }).join('');
 }
 
+function generateTypings(basePath, modules, components) {
+  const importModules = modules.map((f7Module) => {
+    const capitalized = capitalize(f7Module);
+    return `import {${capitalized} as ${capitalized}Namespace} from '${basePath}/modules/${f7Module}/${f7Module}';`;
+  });
+
+  const importComponents = components.map((component) => {
+    const capitalized = capitalize(component);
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks! Merged

PR #3009, opened 2019-02-12
let --f7-message-sent-bg-color being set even with --f7-theme-color

GuillaumeBiton:
--f7-theme-color is a fallback if --f7-message-sent-bg-color isn't set


nolimits4web:
Thanks

GuillaumeBiton:
sorry for the mistake it should be on message-sent bubble.

PR #3032, opened 2019-02-25
Update input.js

bumprat:
Type error, cause textarea unable to fire textarea:resize event.


DAnn2012:
Hi @bumprat 

I make you notice that the changes must be made especially in this file:

` framework7/src/phenome/components/input.jsx`

Thanks.

PR #3033, opened 2019-02-25
Update input.jsx

bumprat:
Typing error causes <f7-input type="textarea" @textarea:resize> unable to fire textarea:resize event.


nolimits4web:
Thanks!

Issue #3073, opened 2019-03-14
vdom - Inline SVG compatibility (case sensitive)


spoxies:
I just learned (the hard way), that SVG elements are case-sensitive. The SVG spec says; SVG is XML and is case-sensitive (unlike HTML).

In F7 the tags are made lower-case (if included as a component), and therefore those lower cased nodes will not function. 
```
<animateTransform/>, <linearGradient/>, <feComponentTransfer/>, <feMerge/>, ...etc
```

**Example:**
```
<template>
   <div class="page" data-name="..">
      <svg xmlns=".." ..>
         <defs>
            <linearGradient ..>
               <stop offset="5%" .. />
            </linearGradient>
         </defs>
         <animateTransform fill="freeze" .. />
         <path .. />
      </svg>
   </div>
</template>
```
**_would become:_**
```
<template>
   <div class="page" data-name="..">
      <svg xmlns=".." ..>
         <defs>
            <lineargradient ..>
               <stop offset="5%" .. />
            </lineargradient>
         </defs>
         <animatetransform fill="freeze" .. />
         <path .. />
      </svg>
   </div>
</template>
```
I will be submitting a PR that checks if an element is an instance of SVGElement, before making it lower-case. It would probably slightly decrease the performance/speed. Please see if you think it would be worth it.

Many thanks for this great project there is nothing more to wish for, it is stunningly complete!!

nolimits4web:
Fixed by #3074

PR #3074, opened 2019-03-14
Feat: Inline SVG compatibility #3073

spoxies:
This PR adds SVG compatibility by checking if an el/node is an instance of SVG.


nolimits4web:
Hi! Thanks, but please no changes in `packages/` files, can you please leave changes only in `src/` file?

nolimits4web:
Great, thanks!

PR #3087, opened 2019-03-20
The list of components in Kitchen Sink is not in alphabetical order

DAnn2012:
The list of components in Kitchen Sink is not in alphabetical order


nolimits4web:
@DAnn2012 same here, let's List will be on top of List Index

nolimits4web:
Thanks!

PR #3112, opened 2019-04-01
Update device.js for iPod touch detection

DAnn2012:
In Device detection library is missing <code>device.ipod = true</code>


nolimits4web:
Thanks!

PR #3177, opened 2019-05-16
Update smart-select-class.js

zhennann:
`optionEl` maybe not found

### DIFF
--- a/src/core/components/smart-select/smart-select-class.js
+++ b/src/core/components/smart-select/smart-select-class.js
@@ -169,9 +169,11 @@ class SmartSelect extends Framework7Class {
       }
     } else {
       optionEl = ss.$selectEl.find(`option[value="${newValue}"]`)[0];
-      displayAs = optionEl.dataset ? optionEl.dataset.displayAs : $(optionEl).data('display-as');
-      text = displayAs && typeof displayAs !== 'undefined' ? displayAs : optionEl.textContent;
-      optionText = [text];
+      if (optionEl) {
+        displayAs = optionEl.dataset ? optionEl.dataset.displayAs : $(optionEl).data('display-as');
+        text = displayAs && typeof displayAs !== 'undefined' ? displayAs : optionEl.textContent;
+        optionText = [text];
+      }
       ss.selectEl.value = newValue;
     }
     ss.$valueEl.text(optionText.join(', '));


nolimits4web:
Thanks!

ringsatu:
I'm having this problem too, below is my code 

```
<f7-list-item title="Kategori" smart-select :smart-select-params="{openIn: 'popup', closeOnSelect: true}">
				<select 
					ref="category" 
					name="category" 
					:value="residentReportForm.category"
					@change="residentReportForm.category = $event.target.value">
					<option
						v-for="item in residentReportCategories" 
						:value="item.id"
						v-bind:key="item.id">{{ item.name }}</option>
				</select>
			</f7-list-item>
```

I have to use `openIn: 'popup', ` otherwise if it is as a 'page' it will throw an error:
`Cannot read property 'dataset' of undefined`

zhennann:
@ringsatu 
I have tested your codes, which work well as openIn: 'page'.

ringsatu:
@zhennann thanks for reply, I tried as well, it works if I set timeout until my categories state fully loaded, but I still get this error, it seems `optionEl = ss.$selectEl.find(`option[value="${value}"]`)[0];` always undefined since the select options is not in the DOM yet?

```
smart-select-class.js:156 Uncaught TypeError: Cannot read property 'textContent' of undefined
    at HTMLInputElement.handleInputChange (smart-select-class.js:156)
    at HTMLDivElement.handleLiveEvent (dom7.module.js:392)
```

 it strange, if I open in popup, it works, I don't even have to use timeout

ringsatu:
my other smart select in other page works really well with the same code

zhennann:
Are you sure you use the newest f7 v4?
image

ringsatu:
Yes, Its 4.4.0

Capture

zhennann:
You can try the newest code!
image

ringsatu:
Ok now updated to v4.4.7 before I have 3 smart selects, the first one has issue, now the first and 3rd one has the same issue

zhennann:
May you give me the while code to reproduce the issue?

ringsatu:
you can check here https://drive.google.com/file/d/1ZJ04oIT0pZCodo9V811M_8hEa-J90eHD/view

zhennann:
As the video, the error line maybe not the same code line number as the newest code.
You can check the code of `smart-select-class.js` is the same as you installed.
this path is : `/framework7/packages/core/src/core/components/smart-select/smart-select-class.js`

ringsatu:
it is same error

```
smart-select-class.js:154 Uncaught TypeError: Cannot read property 'dataset' of undefined
    at HTMLInputElement.handleInputChange (smart-select-class.js:154)
    at HTMLDivElement.handleLiveEvent (dom7.module.js:392)
```

ringsatu:
@nolimits4web any solution for this?

PR #3207, opened 2019-06-11
Close panel if opened-prop is set to false

### LINKED ISSUE #3206: Panels opened property can't be used to close panel
* Framework7 version: [e.g. 4.4.3]
* Vue.js version: [e.g. 2.6.10]
* https://jsfiddle.net/ybxqogre/7/

### Description
If you want to control a panel using the `opened` property, F7 always opens the panel even is the property is set to false. (Have a look at the fiddle above) 

Following code is the reason for this behavior:

```
      if (opened) {
        self.$f7.panel.open(side);
      } else {
        self.$f7.panel.open(side); // <-- should be close!
      }
```
https://github.com/framework7io/framework7/blob/master/packages/vue/components/panel.js#L83

### Expected behavior
If the `opened` property is set to false f7 should close the panel.

### Actual Behavior
F7 always opens the panel when the `opened` property has changed.

I also prepared a pull request to fix this issue which will follow shortly.

wesionaire:
Fixes #3206

### DIFF
--- a/src/phenome/components/panel.jsx
+++ b/src/phenome/components/panel.jsx
@@ -82,7 +82,7 @@ export default {
       if (opened) {
         self.$f7.panel.open(side);
       } else {
-        self.$f7.panel.open(side);
+        self.$f7.panel.close(side);
       }
     },
   },


nolimits4web:
Thanks, but any fixes should be done in src files `src/phenome/components/panel.jsx` not in package. Can you change it there instead?

wesionaire:
But this is a F7 **vue** specific issue. It does **not** affect the core library. I can't find any vue related code inside the src folder. In ``src/phenome/components/panel.jsx`` there are no vue watch methods. The  framework7-vue repository also looks not like the right place as it says "deprecated" on top...

nolimits4web:
This file `src/phenome/components/panel.jsx` is generated to that Vue component file

wesionaire:
Ah I see :+1: . Alright changed it inside the panel.jsx

nolimits4web:
Thanks! 👍

PR #3263, opened 2019-07-21
Update pull-to-refresh.html

DAnn2012:
Excluding the number 0 from the range of randomly generated numbers for cover images


nolimits4web:
Thank you 👍

PR #3267, opened 2019-07-23
Update tooltip.vue

zhennann:
1. tooltip: change the title
2. infinite-scroll / pull-to-refresh: `tabEl` maybe also is `page-content`


nolimits4web:
Merged, thanks 👍

PR #3369, opened 2019-10-25
Style month input field to be consistent with other date fields

edwinnyawoli:
This patch applies the same styling the other date, time etc input fields have to the **month** field so everything looks uniform


nolimits4web:
Thanks!

PR #3427, opened 2019-11-28
Added an alternative by localStorage when isn't available

eem-dev:


shastox:
_localStorage when isn't available_  When can it be?

eem-dev:
Hi @shastox,

There are several situations that can make localStorage unavailable. Some of them are well explained here: https://michalzalecki.com/why-using-localStorage-directly-is-a-bad-idea/

In our case, the problem appeared in browsers of older versions of Android. But we also had reports of users using Chrome in corporate environments, which have domain security rules that prevent localStorage from being used.

The change we are proposing is an alternative to these cases which further increases f7 compatibility.

shastox:
F7 v5 works on iOS >=12 and Android 5. LocalStorage works fine. The cases described are exceptions, you can limit anything and everything.

In any case, replacing localStorage by memory storage is a bad idea

eem-dev:
Hi @shastox,

We are using v3 precisely to provide compatibility with older versions of iOS and Android.

The change we are suggesting is for v3 only.

I agree that localStorage works very well in most cases and devices.

However, we tested our webapp built with F7 v3 running on Android 5+ and due to the particulars of the apps running our webapp, localStorage was not available.

nolimits4web:
Sorry, v3 is not maintained anymore, so no more v3 releases will happen. You can just add this fix to your local version of F7 v3

PR #3435, opened 2019-12-04
Fix navbar no-shadow

NikitaKA:
This PR fixes `no-shadow` class of the navbar component.


NikitaKA:
I found that there is another similar issue with `no-hairline` class. Maybe this PR is not suitable...

nolimits4web:
Yeah, these classes should be applied to main navbar element. Also PR should be submitted to `/src/` folder only. No changes to `/packages` should be made

Issue #3448, opened 2019-12-14
removePage in components-rotuter.js causes crash when $pageEl length is zero


orhan-swe:
* Framework7 version: master
* React.js version: 16
* Platform and Target: Chrome

This sometimes happens on page reload. removePage function in components-rotuter.js causes crash when $pageEl length is zero:

![image](https://user-images.githubusercontent.com/10519489/70855145-22894200-1ec6-11ea-8b25-da8d2e65208e.png)

to fix this we should also check if $pageEl length is zero

PR #3449, opened 2019-12-14
fix for #3448 - removePage in components-rotuter.js causes crash when…

orhan-swe:
… $pageEl length is zero

A check to fix:
Uncaught TypeError: Cannot read property 'f7Page' of undefined

### DIFF
--- a/packages/react/utils/components-router.js
+++ b/packages/react/utils/components-router.js
@@ -59,7 +59,7 @@ export default {
       viewRouter.setPages(viewRouter.pages);
     },
     removePage($pageEl) {
-      if (!$pageEl) return;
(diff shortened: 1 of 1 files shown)


DAnn2012:
Hi @orhan-swe 
Waiting for @nolimits4web  to decide to approve your change, I want to warn you that the changes should be submitted to `/src/` folder only. No changes to `/packages` should be made.

The change should be made to this file:

https://github.com/framework7io/framework7/blob/142f2d6936a546670244446403e65216faf46c1c/src/phenome/utils/components-router.js#L62

Thanks.

nolimits4web:
Thanks, fixed in https://github.com/framework7io/framework7/commit/2e707911ba8e452b538f379da25d3e81cc473277

PR #3450, opened 2019-12-15
make the error explicit when a wrong selector is used to create a view

paulovieira:
This is a simple change that gives a better error message for the case when the user tries to create a view like so:

```js
let myView = app.views.create('.somethin');
```

and there is no element matched by `.somethin` (usually this is a simple typo).

Currently F7 gives a cryptic error with this message:

```
Uncaught TypeError: Cannot set property 'f7View' of undefined
```

(which is clear for anyone who knows the internals of F7, but not at all for users who are new to it)

### DIFF
--- a/src/core/components/view/view-class.js
+++ b/src/core/components/view/view-class.js
@@ -17,10 +17,10 @@ class View extends Framework7Class {
     };
 
     if ($el.length === 0) {
-      let message = `Framework7: can't create a View instance because `;
-      message += (typeof el === 'string') ? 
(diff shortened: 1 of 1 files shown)


nolimits4web:
Thanks! But can you also add another error with `else if ($el.length === 0)`? Because in this case it won't be created as well, with some generic error message like `Framework7: can't create a View instance because passed View element is wrong` or something like that

nolimits4web:
Thanks, merged! 👍

Issue #3488, opened 2020-01-27
Feature request: Custom picker `scrollToEl`


dtom90:
**Is your feature request related to a problem? Please describe.**
By default, when a picker is opened, the screen scrolls the input element of the picker into view:

<img width="300" alt="Scroll to input" src="https://user-images.githubusercontent.com/911671/73209898-29d37a80-4117-11ea-9030-9bfc53080c09.png">

However, some users may want the screen to scroll a different element into view, such as the "Enter" button in this example:

<img width="300" alt="Scroll to button" src="https://user-images.githubusercontent.com/911671/73209906-2e982e80-4117-11ea-9f14-ce7a00a2438c.png">

**Describe the solution you'd like**
I'd like to add an optional picker parameter `scrollToEl` which overrides `inputEl` as the element that will be scrolled to upon opening the picker.

**Describe alternatives you've considered**
Implementing custom code outside Framework7 for scrolling to the correct element. However, this would be redundant and more complex.

nolimits4web:
Fixed with #3489

PR #3489, opened 2020-01-27
Feature: Custom picker `scrollToEl`

dtom90:
Addresses #3488

Adds optional picker parameter `scrollToEl` which overrides `inputEl` as the element that will be scrolled to upon opening the picker.


nolimits4web:
Thanks!

PR #3534, opened 2020-02-21
add css variable for the additional page transitions (parallax)

paulovieira:
`transition-duration` can be set by the user for the default transition (using `--f7-page-transition-duration`), but for the additional transitions (cover, parallax, etc) this is hard-coded.

This PR is done only for parallax because I want to know if there is interest in adding these new variables. If so, I'll update the PR for the remaining effects.


nolimits4web:
Merged, thanks! 👍

PR #3536, opened 2020-02-22
Shouldn't set to opened if loadChildren error

zhennann:


nolimits4web:
Thanks! But i've added that feature in a but more easier way)

PR #3549, opened 2020-02-29
Default size for img used as notification's icon

Pitouli:
When using an img as an icon for notification (something officially supported since it's described in the doc), the size of the image is not adapted. It's easy to correct, but it would be even more easier to have it working out of the box.


nolimits4web:
Merged, thanks!

Issue #3577, opened 2020-03-16
Messagebar send-link slot is not rendered


AliMamed:
* Framework7 version: 5.5.0
* React.js version: 16.13.0
* Platform and Target: any
* Live Link or JSFiddle: https://jsfiddle.net/aqo6d1e8/

### Describe the bug
Messagebar's send-link slot contents is not rendered
IMHO problem is here https://github.com/framework7io/framework7/blob/master/src/phenome/components/messagebar.jsx#L147 when it is converted to https://github.com/framework7io/framework7/blob/master/packages/react/components/messagebar.js#L213
as `F7Link` receives the element also to 'send-link' slot and doesn't render it as it doesn't has such slot

nolimits4web:
Fixed by #3578

PR #3578, opened 2020-03-16
Copy children with slot prop other than `default` to `default`

AliMamed:
This change is proposed to resolve Issue #3577


nolimits4web:
Merged, thanks! 👍

PR #3580, opened 2020-03-18
Update actions-class.js

zhennann:
maybe more extensible


nolimits4web:
No, this is not needed. Actions already has same events as popover defined in popover's `on` settings

zhennann:
If Actions showing as popover, the actions instance will donot known when the popover opened/closed. So need to emit the popoverOpened/Closed on Actions.

nolimits4web:
These events already emitted here https://github.com/framework7io/framework7/blob/master/src/core/components/actions/actions-class.js#L108-L124

zhennann:
wonderful，i see that

Issue #3617, opened 2020-04-28
[Bug ? Feature Request ?] Texteditor-Popover is incorrectly positioned


Pitouli:
* Framework7 version: latest
* Vue.js version: latest
* Platform and Target: all
* Live Link or JSFiddle: https://jsfiddle.net/y95raxdj/

### Describe the bug
The css position of the texteditor-popover is relative to the `.framework7-root` element, but it seems it is calculated relative to the `html` element. 
Therefore, If the dev limit the size of the `body` element (for example on big screens), then the popover is incorrectly positioned  

### To Reproduce
1. Go to https://jsfiddle.net/y95raxdj/
2. Type then select text to make the popover appears
3. Observe that it is offset as much as the red padding

### Expected behavior
Since the popover is relative to the `.framework7-root` element, the `top` and `left` position should be computed accordingly.
Note : even better (but more complex), the popover should be relative to the text-editor element, so it scrolls with it.

### Actual Behavior
`top` and `left` are computed relative to `html` element.

### Additional context
It's not really a bug since resizing the app is not part of Framework7.
But I believe it's a common need (on 21:9 screens, having the app fullwidth is not very elegant), and since it's not very hard (substract the position of the .framework7-root to the element coordinates), it could be a good improvement! :)

Pitouli:
Since I'm opening thousands of tickets, and this one is a modification I'm skilled enough to do, I thought it would be better to write the PR myself!

nolimits4web:
Fixed by #3618

PR #3618, opened 2020-04-28
text-editor-popover position relative to f7-root

Pitouli:
As described here: https://github.com/framework7io/framework7/issues/3617
The popover is relative (css'ly speaking) to the #framework7-root element, but is computed relative to the html element.
The proposal is to compute the position relative to the #framework7-root element

**Important: I haven't been able to test if for real! It should work, but it should definitely not be merged without real verification!**


Pitouli:
I just remembered that I HAD a Framework7 running on which testing my code... On the node_modules of my own project. 
So I tested, and it works as expected (a review is still welcome of course).

nolimits4web:
Thanks, merged!

PR #3630, opened 2020-05-04
Panel transition when pushing popup opens

Pitouli:
My app is using colors for headers, and I observed that panels do not behave beautifully when a link inside the panel is opening a "pushing" popup _(it's also the case with default colors, but it's less obvious, at least with the dark theme on revealed panels, since the header of the panel is almost black)_

The PR implements some transitions that make the app beautiful again.

It takes care of right and left panels, different effects for cover and reveal panel (respectively "go away" and "go below" effect), and works well when the panel is closed at the same time that it opens the popup. 

Few things to notice:

- I changed the transition duration so it matches the one of the popup: so when the popup is closing, the panels and the pushed view are back to their respective place at the same time
- It does not take care of "progressively going back" when the user is swiping to close: it would imply to also change the "pushed view" behavior so it moves horizontally and not only vertically. But it is definitely not a requirement ; as is, it's already very good looking.
- The `translate3d(calc((1 - var(--f7-popup-push-scale, 1)) * 50vw), 0, 0px)` is required because the scale alone keeps the panel visible _(since the panel is less wide than the view, scaling it down of the same amount make it "translate" less)_. 
The formula ensures that the panel is translated left as much as the visible "margin" between the pushed view and the viewport created by the scaling of the view _(I consider here that "pusing popup" are pushing only on screens where `f7-root.width === viewport.width`)._ And then, since the panel is scaled down too, it's even more "away" from the edge of the viewport.
An alternative is to change the origin of the scale (see next comment)
- I initially added border radius for the "reveal" panels _(must be done two times, on the panel itself and on the view inside)_, but it is not necessary: with the "double" translation (see previous point), the corners are not visible anyway _(unless your panel is extremely skinny, or your radius extremely big)_

You can test everything on this fiddle: 
https://jsfiddle.net/eazLq8ku/3/

### DIFF
--- a/src/core/components/panel/panel.less
+++ b/src/core/components/panel/panel.less
@@ -186,6 +186,30 @@ html {
     transform: translate3d(0px,0,0);
   }
 }
+
+html.with-modal-popup-push .framework7-root > .panel {
+  &.panel-in,
+  &.panel-out {
+    transition-duration: var(--f7-popup-transition-duration);
+
+    &.panel-reveal.panel-left {
+      transform: translate3d(calc((1 - var(--f7-popup-push-scale, 1)) * 50vw), 0, 0px) scale(var(--f7-popup-push-scale, 1));
+    }
+    &.panel-reveal.panel-right {
+      transform: translate3d(calc((1 - var(--f7-popup-push-scale, 1)) * -50vw), 0, 0px) scale(var(--f7-popup-push-scale, 1));
+    }
+    &.panel-cover.panel-left {
+      transform: translate3d(-100%, 0, 0px);
+    }
+    &.panel-cover.panel-right {
+      transform: translate3d(100%, 0, 0px);
+    }
+  }
+}
+html.with-modal-popup-push-closing .framework7-root > .panel.panel-in {
+  transition-duration: var(--f7-popup-transition-duration);
+}
+
 .panel-resize-handler {
   position: absolute;
   top: 0;


Pitouli:
An alternative by replacing the "complex" translation by a simple "transform-origin" for the scale. 

https://jsfiddle.net/eazLq8ku/7/

The "issue" is that now, when the viewport is equal to the f7-root, the panel is perfectly aligned with the pushed view. So you have to take care of the border radius.

The code is simpler to understand, but it is longer (and the generated code is even longer).
But it removes the `calc()` that can be expensive (only keep one for right panel). And in terms of UX, it makes the panel animation start and end exactly at the same time than the popup (we are talking of a 50ms difference, so I'm not even sure the user can perceive it...)

Anyway, if this second version is better, I will be glad to update my PR :) 

```
html.with-modal-popup-push .framework7-root > .panel {
  &.panel-in,
  &.panel-out {
    transition-duration: var(--f7-popup-transition-duration);
    &.panel-reveal.panel-left {
      transform: scale(var(--f7-popup-push-scale, 1));
      transform-origin: 50vw center;
      background: black;
      .view {
        border-top-left-radius: var(--f7-popup-push-border-radius);
      }
    }
    &.panel-reveal.panel-right {
      transform: scale(var(--f7-popup-push-scale, 1));
      transform-origin: calc(var(--f7-panel-width) - 50vw) center;
      background: black;
      .view {
        border-top-right-radius: var(--f7-popup-push-border-radius);
      }
    }
    &.panel-cover.panel-left {
      transform: translate3d(-100%, 0, 0px);
    }
    &.panel-cover.panel-right {
      transform: translate3d(100%, 0, 0px);
    }
  }
}
html.with-modal-popup-push-closing .framework7-root > .panel.panel-in {
  transition-duration: var(--f7-popup-transition-duration);
  &.panel-reveal.panel-left {
    transform-origin: 50vw center;
  }
  &.panel-reveal.panel-right {
    transform-origin: calc(var(--f7-panel-width) - 50vw) center;
  }
}
```

nolimits4web:
Thanks! 👍  I decided to go with your 2nd option + added same for Sheet-Push as well

Pitouli:
@nolimits4web : small bug: when the panel is open, you click a link to open a pushing popup, then you close the popup, the panel is still open (as expected), but the "main view" (which overlap the panel [i'm testing with reveal effect]) keeps the rounded corners it had when being pushed (easier to see if the header is colored), which makes white corners.

nolimits4web:
@Pitouli can you create a new issue with demo fiddle of the issue, please? :)

Pitouli:
@nolimits4web sadly, the bug cannot be reproduced on jsfiddle. 
It appears only on the iPhone.

So here is the jsfiddle: https://jsfiddle.net/unes52o3/1/
And here is an export of a jsfiddle I hosted on my website you must access from an iPhone to reproduce the error: bug-panel-popup-framework7.htm

You should observe a top left white corner on the main view after having opened then closed the popup with the link in the left panel.

Issue #3659, opened 2020-06-12
page:afterout event skipped if exiting route calls router.navigate to a component destination


peitschie:
* Framework7 version: 5.7.7
* Platform and Target: N/A
* Live Link or JSFiddle: https://jsfiddle.net/qmxnb2gL/2/

### Describe the bug
When navigating between component-based pages, if the component page that is exiting accidentally performs another `router.navigate`, the `page:afterout` event is never triggered (though `page:beforeout` **is** still triggered correctly).

This bug does not happen if using a `content` destination for some reason... only when using a `component` destination.

### To Reproduce
See JSFiddle at https://jsfiddle.net/qmxnb2gL/2/. The basic steps are:
In the JSFiddle, note that the top link shows two dialogs, for both `page:beforeout` and `page:afterout` events.

The second link exhibits the bug, where only a single dialog is shown during the transition for `page:beforeout`.

In addition, after using the second link, the "Back" link is broken.

### Expected behavior
I expect the `page:afterout` event to be still fired. The Back link should still function to do something. Should the second `router.navigate` instruction also be ignored? I don't really know...

### Actual Behavior
No `page:afterout` event is fired when navigate is accidentally called twice. Router's back link then breaks.

### Additional context
In my real application, this is caused by a chain of events that accidentally decides to navigate to the same route the application had already decided to navigate to. It's clear that my application shouldn't be calling `navigate` twice in these circumstances... but the Framework7 behaviour was sufficiently odd (the second navigate instruction isn't ignored, and breaks the `page:afterout`) that it seemed worth raising.

PR #3668, opened 2020-06-24
Core: Router - lock router on component load

peitschie:
Replicating fix from 9488d98ea0b4663fbf7fc9954a0e242918a91d5b for back

Related to #3659


nolimits4web:
Merged, thanks!

PR #3681, opened 2020-07-21
enable bind:value for svelte

tonprince:
I would like to request to add bind:value support for text and other input types:

Issue: https://forum.framework7.io/t/how-to-bind-a-value-to-svelte-input/9656


amit777:
It seems like this change set breaks a lot of the coding conversions. Lots of double quotes instead of single quotes for example.

I too would like the components to support bind syntax.

nolimits4web:
closing as it is supported in v6

Issue #3696, opened 2020-08-20
Open panel in defined container


AliMamed:
Currently panel is mounted to `app.root`.
I propose to have a parameter to be able to mount panel to some other container.
For example to mount it inside a modal popup.

I will make a PR.

PR #3697, opened 2020-08-20
Panel container element parameter

AliMamed:
adds parameter to be able to mount panel to defined container element 
issue #3696


nolimits4web:
Merged, thanks!

PR #3703, opened 2020-08-27
Fix formatting for some supporters

theryan722:
Minor formatting corrections in the backers file to be in line with the rest of the document


nolimits4web:
Thanks

PR #3707, opened 2020-09-05
Update actions.d.ts and sheet.svelte

DAnn2012:
Fixed Typos

(Also in component.d.ts, app-class.d.ts and view.d.ts)


nolimits4web:
Thanks!

PR #3708, opened 2020-09-06
Fixed issue template name field

DAnn2012:
Fixed issue template name field _Framewor7_ -> **Framework7** for React, Svelte and Vue (The one in the Core version was fixed already six months ago).

I would like to point out that the same typo is also present in the name of these four files:

file-names


nolimits4web:
Thanks!

PR #3711, opened 2020-09-09
Correct text of the "Password" dialog example

DAnn2012:
for Kitchen Sink Core / React / Svelte / Vue

A typo on "list" pages is also fixed.


nolimits4web:
Merged, thanks!

PR #3718, opened 2020-09-24
Picker does not have an addValue method

DAnn2012:
On the "Picker" page of the documentation, there is also addValue in the method list:

https://github.com/framework7io/framework7-website/blob/2f3a2d71f8e7898dca2c7d9b669e86c53a87afd1/src/pug/docs/picker.pug#L347-L349

But this method doesn't exist in

https://github.com/framework7io/framework7/blob/master/src/core/components/picker/picker-class.js


nolimits4web:
Thanks!

PR #3731, opened 2020-10-02
ListInput in Svelte does not have a defaultValue property

DAnn2012:
Update form-storage.svelte


nolimits4web:
Merged, thanks! @DAnn2012 for future KS PRs please send them into v6 branch instead of master. Now merging from master to v6 KS is almost not mergeable

PR #3732, opened 2020-10-02
[v6] Fixed typos in CHANGELOG.md

DAnn2012:
Update CHANGELOG.md


DAnn2012:
- Lite version doesn't contain
      - Template7

@nolimits4web Does this reference to **Template7** have to be made here even if it has been completely removed?

nolimits4web:
Thanks! Yes, Template7 statement was added when i thought it will still be present in core version)

PR #3736, opened 2020-10-07
Added Aurora in app-class.d.ts

DAnn2012:
Update app-class.d.ts, framework7-phenome.d.ts, framework7-react.d.ts and framework7-vue.d.ts


nolimits4web:
Merged, thanks!

PR #3740, opened 2020-10-08
Fixed typo in router.d.ts

DAnn2012:
Update router.d.ts

Also fixed typo in Photo Browser KS pages (Core/React/Svelte/Vue), request.js and request.d.ts

### DIFF
--- a/kitchen-sink/core/pages/photo-browser.html
+++ b/kitchen-sink/core/pages/photo-browser.html
@@ -33,7 +33,7 @@
         </div>
       </div>
       <div class="block block-strong">
-        <p>Photo Browser suppots 2 default themes - default Light (like in previous examples) and Dark theme. Here is a
+        <p>Photo Browser supports 2 default themes - default Light (like in previous examples) and Dark theme. Here is a
           Dark theme examples:</p>
         <div class="row">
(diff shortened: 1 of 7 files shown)


DAnn2012:
https://github.com/framework7io/framework7/blob/69a577c97af9f16d4943a9f4b891cb017eacd3cf/src/core/shared/request.js#L324

@nolimits4web  This **message: xhr.status** is right or should it be **message: xhr.statusText**?

nolimits4web:
@DAnn2012 it was `xhr.status` in before (in v5) but i think `.statusText` has more sense here

DAnn2012:
@nolimits4web So will you make the change after merging this PR?

I also point out that there are some references to **Template7** in this file:

https://github.com/framework7io/framework7/blob/43480d032de6a4804c562ebe1d6fc767f529aca3/kitchen-sink/core/pages/page-loader-component.html#L17-L20

https://github.com/framework7io/framework7/blob/43480d032de6a4804c562ebe1d6fc767f529aca3/kitchen-sink/core/pages/page-loader-component.html#L38

nolimits4web:
Thanks. I've fixed mentioned issue

PR #1386, opened 2017-03-02
Fix issue #1385

Tarekajaj:

Issue #1385, opened 2017-03-02
Wrong range selected in the datepicker


tatxini:
This is a (multiple allowed):
* [x] bug

* Framework7 Version: 1.5.3
* Platform and Target: Material theme & iOS Theme.

### What you did
Select a date from range datepicker

### Expected Behavior
When you select the range, all days shoud be selected in the calendar.

### Actual Behavior
The last day on some cases is not marked, but the date above is correct.

This issue is reproducible in the kitchen sink, both material theme and iOS theme.
The gif shows some examples with the wrong behavior, and the last one is the expected behaviour.

![bug range calendar f7](https://cloud.githubusercontent.com/assets/25824042/23506536/1045c496-ff49-11e6-91b5-6813ff4df309.gif)

nolimits4web:
Well, for some reason i can not replicate it and it works fine for me. Can anyone confirm there is such issue cc @valnub @ZanderBrown

valnub:
I can't replicate it in the online Kitchen Sink demo at http://framework7.io/kitchen-sink-material/ using Google Chrome on latest macOS.

@tatxini Which browser and OS are you using?

valnub:
Yes, I can replicate it using Google Chrome on macOS.

The problem only occurs if you select exactly the same date range as in the gif (March 16th-30th).

On other date it works correctly.

valnub:
I also see another problem in the date picker. When I click on a day, it will not always select the day (nothing happens). Sometimes I have to click multiple times on a day and then it will select it. I remember that someone posted about a change in Google Chrome that breaks clicking behavior in F7. Don't remember that post tho. This problem does not exist in other browsers, e.g. Firefox. Other users also reported similar issues with components not always reacting when clicked.

macdonst:
I can reproduce it as well but I haven't discovered what the magic combination is. The first few date ranges I select work fine then eventually the date range stops highlighting the last day.

nolimits4web:
Still works fine for me 😞 What is environment. I'm checking at:
http://framework7.io/kitchen-sink-material/
macOS 10.12.3
Chrome OSX 56.0.2924.87 (64-bit) (with both emulating mobile device (android) and not)

Tarekajaj:
Was about to post about this issue, Happens for me on windows 7 64bits, chrome 56.0.2924.87 and on firefox 51.0.1. Also I can reproduce it on iPad pro v10.2 & a Samsung Note 5 running android 6.0.1

[...]

PR #3751, opened 2020-10-28
Fixed wrong range DST dates selected in the datepicker

DAnn2012:
Fixed an old issue described in #1385 (daylight saving dates in the selected range).

I sent a solution only for the next version 6, because I think there will be no further updates to version 5, but the problem affects all previous versions of f7 as the solution that was proposed in #1386 caused problems with some time zones and is been reverted in https://github.com/framework7io/framework7/commit/8c9064fcfbd1bec65da8c4977cd483c57e891cb1.


nolimits4web:
Merged, thanks!

Issue #3754, opened 2020-10-29
Support for React v17


fny:
* Framework7 version: 5.7.13
* React.js version: 17.0.1
* Platform and Target: Not releveant
* Live Link or JSFiddle: Not relevant

### Describe the bug

Framework 7 has peer dependency  "react": "16.x" this should be expanded to React v17.

React v17 is intended to be a drop in replacement, so there are no breaking changes.

PR #3756, opened 2020-10-30
Add React v17 as peer dependency

fny:
This addresses #3754 for F7 v5.

React v17 is backwards compatible, so this should help anyone that wants to upgrade.


nolimits4web:
Thanks!

PR #3758, opened 2020-11-07
document backdropEl as parameter

stevenroussey-privicy:
You added this as a parameter to change globally. This only changes for popup. Is it available for others?


nolimits4web:
Thanks, I pushed the correct fix to all related components

PR #3760, opened 2020-11-08
[v5 only] Backport of my most recent fixes to the KS files

DAnn2012:
I saw that the master branch is still receiving changes for maintenance, so I thought I'd backport my most recent fixes that I already posted some days ago and that have been included in v6 so that @nolimits4web  can evaluate if that's the case for also include them in v5.


nolimits4web:
Thanks!

PR #3775, opened 2020-11-23
Fixed typo in panel.d.ts

DAnn2012:
Update panel.d.ts


DAnn2012:
@nolimits4web In this line:

https://github.com/framework7io/framework7/blob/0cad54bc10530301bd71baf2b8f7a97d7f13ed92/src/core/components/panel/panel.d.ts#L85-L86

the type `HTMLElement | CSSSelector` is correct or it should be `Dom7Instance`?

nolimits4web:
It should be `Dom7Array`, fixed, thanks!

PR #3782, opened 2020-12-02
Core: Vdom - fix custom component in template7 each helper err

insomenia-seongjun:
I think there is an error in the vnode.
An error occurred when I wrote the code as follows. ( in `blahblah.f7.html`)
```html
<template>
 ...
{{#each items}}
  <custom-component></custom-compoent>
{{/each}}
...
</template>
```

It's a problem when you leave the page where the above code is written.

image

I think an exceptional situation occurs when the custom components are in template7 each helper when they are deleted and created.

### DIFF
--- a/src/core/modules/component/vdom.js
+++ b/src/core/modules/component/vdom.js
@@ -71,7 +71,7 @@ const destroyCustomComponent = (vnode) => {
   // eslint-disable-next-line
   const component = vnode && vnode.elm && vnode.elm.__component__;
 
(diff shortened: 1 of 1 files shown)


insomenia-seongjun:
I found an error condition.
If the Custom Component is in the tab class, The error will occur
```
{{#item}
<div class='tab' id='first-tab'></div>

 <custom-component> </custom-componet>
{{/item}}

```

Issue #3788, opened 2020-12-22
props configured in route options not passed to react component on initial load


usrlocl:
* Framework7 version: 6.0.0-beta.4
* React.js version: 17.0.0
* Platform and Target: Web
* Live Link or JSFiddle: NA

### Describe the bug
props configured in route options are not passed to react component on the initial route. They are however passed on subsequent navigation (by clicking link).

### To Reproduce
1. Configure props on initial route
```
    {
        path: '/',
        component: Home,
        options: {
           props: {
           foo: 'bar',
        }
     }

```
2. Load the main view with root url and push state enabled
```
   <View
        main
        url="/"
        browserHistory
        browserHistorySeparator=""
        browserHistoryRoot=""
      />

```
3. Access props in the Home component
```
    (props) => {
        console.log(props.foo);    // undefined on first load
    }

```
4. props.foo is undefined on first load
5. Navigate away and back to root.
6. props.foo is now available.

### Expected behavior
props should be available on each load of view.

usrlocl:
Hi, I've raised a PR that fixes the issue on my local setup. Unfortunately I couldn't get a fiddle working against the exact scenario. It's a small fix though. Hopefully it should be obvious.

https://github.com/framework7io/framework7/pull/3789

nolimits4web:
Fixed in #3789

PR #3789, opened 2020-12-22
#3788 FIX: [react] pass props from route options for initial route

usrlocl:
getRouterInitialComponent() function sets up components for the default route.
Currently, props configured on the route options are not passed when creating
the initial component, so are not available on first render. (props are
available only on subsequent navigation). This change fixes the issue by copying
props from route options when setting up initial component

Always follow the [contribution guidelines](https://github.com/framework7io/framework7/blob/master/.github/CONTRIBUTING.md) when submitting a pull request.


nolimits4web:
Merged, thanks!

PR #3810, opened 2021-01-14
title -> footer typofix

kozmanbalint:
A small typofix to render footer content to card footer placeholder (instead of title) in f7-vue.

Thanks,
Bálint


nolimits4web:
Merged, thanks!

Issue #3820, opened 2021-02-06
Wrong typings path


sinapcs:
* Framework7 version: 6.0.7
* Platform and Target: Doesn't matter

### Describe the bug
When you want to import ES Modules in a typescript based project you get a missing type definition error.

### To Reproduce
1. start a new Create React App project with TypeScript
```
npx create-react-app my-app --template typescript
```
2. import a component from `framwork7/components`
```js
import Tabs from "framework7/components/tabs";
```
3. See error
`Could not find a declaration file for module 'framework7/components/tabs'. `

### Expected behavior
The `typings` value in `package.json` of a component should look like this:
```json
  "typings": "../../types/components/component_name/component_name.d.ts"
```

### Actual Behavior
The current `typings` value is:
```json
  "typings": "../../types/components/component_name.d.ts"
```
which is missing the `component_name` directory.

nolimits4web:
Fixed by #3821

PR #3821, opened 2021-02-06
add module/component directory to typings path in package.json files

sinapcs:
#3820 
These changes will fix the type declaration paths in the components package.json file.


nolimits4web:
Merged, thanks!

Issue #3884, opened 2021-05-09
f7.actions.create() with multiple button groups gives TypeScript error


D-Marc1:
* Framework7 version: 6.0.17
* Platform and Target: browser, iOS, Android

### Describe the bug
`Actions.Parameters` type currently only has ` buttons?: Button[]` in `action.d.ts`. I'm able to fix the error it gives by simply editing the file and doing ` buttons?: Button[] | Buttons[][]`, in order to allow multiple group.

Let's assume I use something directly from the docs to create an action sheet programmatically.

```js
f7.actions.create({
  buttons: [
    // First group
    [
      {
         text: 'Do something',
         label: true
       },
       {
          text: 'Button1',
          bold: true
        },
        {
          text: 'Button2',
        }
      ],
      // Second group
      [
        {
          text: 'Cancel',
          color: 'red'
        }
      ]
    ]
});
```

Using the action sheet API with multiple button groups yields the following TypeScript error: 

```
Property 'text' is missing in type '[{ text: string; label: true; }, { text: string; bold: true; }, { text: string; }]' but required in type 'Button'.
```

nolimits4web:
fixed by #3885

PR #3885, opened 2021-05-09
Add typing for button groups on action sheet API

D-Marc1:
Add typing for button groups on programmatically calling action sheet API with `f7.actions.create({})`

Per issue: https://github.com/framework7io/framework7/issues/3884#issue-881514177


nolimits4web:
Merged, thanks!

PR #3896, opened 2021-05-28
fix(toolbar): added missing TS DOM Events definitions

DAnn2012:
Update toolbar.d.ts


DAnn2012:
@nolimits4web DOM events are also missing in these files:

- `accordion.d.ts`
- `card.d.ts`
- `grid.d.ts`
- `treeview.d.ts`

PR #3897, opened 2021-05-31
fix(accordion, card, grid, smart-select, treeview): added missing TS DOM Events definitions

DAnn2012:
View https://github.com/framework7io/framework7/pull/3896#issuecomment-850797197

Update accordion.d.ts, card.d.ts, grid.d.ts, smart-select.d.ts and treeview.d.ts


nolimits4web:
Thanks!

PR #3898, opened 2021-06-03
Fix (data-table): added missing TS DOM Events definitions

DAnn2012:
Update data-table.d.ts

@nolimits4web Please, you could take a look at these too: https://github.com/framework7io/framework7-website/issues/502 and https://github.com/framework7io/framework7-website/pull/500

Thanks


nolimits4web:
Thanks!

PR #3943, opened 2021-09-20
Sortable `List` nested `ListInput`'s behavior

forsureitsme:
Bug: `ListInput`'s with `sortable={false}` were rendering the sortable handler when nested inside a sortable list.
This change allows `ListInput`'s sortable value to take precedence over a parent list whenever it's set, otherwise it behaves like usual.


nolimits4web:
Merged, thanks, but instead of nullish coalescing I changed to ternary, as not sure babel is configured to understand it properly

PR #3961, opened 2021-12-09
Fix range: Add missing type definitions

schoero:
Adds missing type definitions for the range component that are documented on the [website](https://framework7.io/docs/range-slider) but not in the type declarations.


nolimits4web:
Merged, thanks!

PR #3972, opened 2022-02-01
fix(types): add missing params to notification and toast open/close methods

peitschie:
These types are not documented on the website, but are supported in code. Both these extend the `Modal` base class, which implements this support for this param.

It's very helpful especially on close to safely and immediately dispose the toast without waiting for the animation to complete.

See:
* https://github.com/framework7io/framework7/blob/master/src/core/components/notification/notification-class.js#L7
* https://github.com/framework7io/framework7/blob/master/src/core/components/toast/toast-class.js#L8
* https://github.com/framework7io/framework7/blob/master/src/core/components/modal/modal-class.js#L73
* https://github.com/framework7io/framework7/blob/master/src/core/components/modal/modal-class.js#L159

### DIFF
--- a/src/core/components/notification/notification.d.ts
+++ b/src/core/components/notification/notification.d.ts
@@ -13,9 +13,9 @@ export namespace Notification {
     params: Parameters;
 
     /** Open notification */
-    open(): Notification;
+    open(animate?: boolean): Notification;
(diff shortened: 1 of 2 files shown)


nolimits4web:
Merged, thanks!

PR #3976, opened 2022-02-08
Do not clone deep options in (router) back

blikblum:
The same as #2739, but for back

Same reason:

In the process of cloning options in back function, the the route instance is replaced by a copy, so the route instance stored in router.currentRoute is different from the one passed to route async handler.

The proposed change is safe since options variable is not mutated at all.

It is also more performant


nolimits4web:
Merged, thanks!

PR #3979, opened 2022-02-08
Do not select stacked pages when querying for previous page in router back

### LINKED ISSUE #3975: Route beforeLeave handler is called twice when using stackPages
* Framework7 version: 6.3.15
* Platform and Target: [Chrome browser]
* Live Link or JSFiddle: https://codepen.io/blikblum/pen/BampNNq

### Describe the bug
When exiting a nested page, `beforeLeave` route handler is called twice, the second time with params different from the actual transition

This occurs when using stackPages: true

### To Reproduce
1. Go to https://codepen.io/blikblum/pen/BampNNq
2. Click on Person link
3. Click on Fact link
4. Click on About link
5. Open console to view beforeLeave logging
6. Click on back button
   beforeLeave is called twice
   first time with from.path === '/about/' and to.path === '/fact/'
   second time is called with from.path === '/fact/' and to.path === '/person/'

### Expected behavior
 beforeLeave is called once with from.path === '/about/' and to.path === '/fact/'

### Actual Behavior
   beforeLeave is called twice
   first time with from.path === '/about/' and to.path === '/fact/'
   second time is called with from.path === '/fact/' and to.path === '/person/'

### Screenshots
[..]

### Additional context
The same behavior is observed in v5

blikblum:
Fixes #3975 

This change does not select .stacked pages when determining previous page.

Uses the same selector used e.g. in https://github.com/framework7io/framework7/blob/master/src/core/modules/router/back.js#L343

### DIFF
--- a/src/core/modules/router/back.js
+++ b/src/core/modules/router/back.js
@@ -734,7 +734,7 @@ function back(...args) {
   }
   let $previousPage = router.$el
     .children('.page-current')
-    .prevAll('.page-previous:not(.page-master)')
+    .prevAll('.page-previous:not(.stacked):not(.page-master)')
     .eq(0);
 
   let skipMaster;


nolimits4web:
Merged, thanks!

blikblum:
Is the preload code path necessary at all when using `stackPages: true`?

A possible fix is to skip preload when stackPages: true

nolimits4web:
I have reverted and pushed another fix, can you check if it works for you now?

blikblum:
Thanks. This fixes the original issue and also prevents the component being recreated when using stackPages.

I noticed the stacked class is kept in previous page when navigating backward giving a different structure when navigating forward in the same page.

In the example when going to the "Fact" page, starting from home, i get the following dom structure:

```html
<div class="view view-main ios-edges">
      <div class="page page-previous stacked" data-name="home" aria-hidden="true"></div>
        
      <div class="page page-previous" data-name="person" aria-hidden="true"></div>
     
      <div class="page page-current" data-name="fact"></div>           
</div>
``` 

When going to "about" page and going back to "fact", i get the following structure (the previous page keeps the stacked class)

```html
<div class="view view-main ios-edges">
      <div class="page page-previous stacked" data-name="home" aria-hidden="true"></div>
        
      <div class="page page-previous stacked" data-name="person" aria-hidden="true"></div>
     
      <div class="page page-current" data-name="fact"></div>           
</div>
``` 

This seems to prevent swipeBack feature.

Should i open a knew issue?

blikblum:
Just found a issue when using stackPages: false with duplicated dom page elements but seems not related to this change because i reverted back and and still persists. Will investigate further and report

PR #3990, opened 2022-02-22
Removed phonegap and added capacitor reference in package.json files

DAnn2012:
Remove phonegap reference from package.json files in packages folders (Core/React/Svelte/Vue)

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -273,7 +273,7 @@
     "iphone",
     "ipad",
     "apple",
-    "phonegap",
+    "capacitor",
     "native",
     "touch",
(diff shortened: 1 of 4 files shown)


nolimits4web:
These are keywords, someone can still search by it, so `phonegap` must be there

DAnn2012:
@nolimits4web 

and what do you think about adding `capacitor` as a keyword?

Thanks.

nolimits4web:
@DAnn2012 I agree with adding `capacitor`

DAnn2012:
@nolimits4web 

Done.

PR #3997, opened 2022-03-03
fix(react): typing error(Actions, Range)

AlwaysLoveme:
src/modules/components/component.d.ts:
```ts
...
 $on: (eventName: string, handler: () => void) => void;  // eventName add type
 $once: (eventName: string, handler: () => void) => void;  // eventName add type
...
```
src/react/components/actions.jsx: `Actions` is namespace, can not use it as type;
src/react/components/range.jsx: `Range.Rage` cause build error

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6129,6 +6129,16 @@
       "license": "ISC",
       "optional": true
     },
+    "node_modules/fsevents/node_modules/ini": {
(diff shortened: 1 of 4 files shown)


nolimits4web:
Hi, there shouldn't be committed changes to package-lock.json. Can you remove them?

PR #4023, opened 2022-05-10
Add export for props interface

Chupriarti:
Adds export for props interface so there will be much easier to add types for custom conponents based on framework7 react components.
Full description here - https://github.com/framework7io/framework7/issues/4024


nolimits4web:
Merged, thank you!

Issue #4055, opened 2022-06-26
@rollup/plugin-typescript TS2307: Cannot find module 'framework7/lite'


kayazinc:
* Framework7 version: 7.0.5
* Svelte version: 7.0.5
* Platform and Target: desktop, nodejs, chrome

### Describe the bug
I cannot reproduce using CodeSandbox.  tbh i have no idea what is wrong since i have issue trying to import Framework7 module:
  
`import Framework7 from 'framework7/lite'`   <--- does not work
`import Framework7 from 'framework7'`  <--- it works

Please advise me, thanks.

Below are the error messages:
---
[0] rollup-plugin-svelte: The following packages did not export their `package.json` file so we could not check the `svelte` field. If you had difficulties importing svelte components from a package, then please contact the author and ask them to export the package.json file.
[0] - framework7
[0] - framework7-svelte

AlwaysLoveme:
Hi, i fixed this error, just waiting for code to merge, or you can copy code to "typesVersions" of package.json in "node_modules/framework7" just for temporarily resolved;

`
{
...

"typesVersions": {
    "*": {
      ".": [
        "./framework7.d.ts"
      ],
      "components/appbar": ["./components/appbar/appbar.d.ts"],
      "components/dialog": ["./components/dialog/dialog.d.ts"],
      "components/popup": ["./components/popup/popup.d.ts"],
      "components/login-screen": ["./components/login-screen/login-screen.d.ts"],
      "components/popover": ["./components/popover/popover.d.ts"],
      "components/actions": ["./components/actions/actions.d.ts"],
      "components/sheet": ["./components/sheet/sheet.d.ts"],
      "components/toast": ["./components/toast/toast.d.ts"],
      "components/preloader": ["./components/preloader/preloader.d.ts"],
      "components/progressbar": ["./components/progressbar/progressbar.d.ts"],
      "components/sortable": ["./components/sortable/sortable.d.ts"],
      "components/swipeout": ["./components/swipeout/swipeout.d.ts"],
      "components/accordion": ["./components/accordion/accordion.d.ts"],
      "components/contacts-list": ["./components/contacts-list/contacts-list.d.ts"],
      "components/virtual-list": ["./components/virtual-list/virtual-list.d.ts"],
      "components/list-index": ["./components/list-index/list-index.d.ts"],
      "components/timeline": ["./components/timeline/timeline.d.ts"],
      "components/tabs": ["./components/tabs/tabs.d.ts"],
[...]

PR #4059, opened 2022-07-07
fix(core): fix core components types file not found in typescript when import

AlwaysLoveme:
#4055 

an error occurred like this：

image

so should add core component types in "typesVersions" of package.json


yoyo930021:
https://github.com/framework7io/framework7/issues/4055#issuecomment-1178495409

Issue #4076, opened 2022-09-20
Router doesn't follow history


Simone4e:
* Framework7 version: 7.0.7
* Platform and Target: Browser + Cordova
* Live Link or JSFiddle: https://jsfiddle.net/r135p4ws/

### Describe the bug

Router back seems like doesn't follow the history although there is the `history: false` option on the middle page (see fiddle)
Scheme of page:
About -> Edit Page (`history: false`) -> newPage

### To Reproduce
Steps to reproduce the behavior (fiddle):
1. Go to page about
2. Click on button and go to edit page
3. Click on button and go to new page
4. Now try to go back you will see that it does not return to the first page

### Expected behavior
It should skip the ignored page with the `history: false` option and go directly to the previous page (as per history).
page -> editPage -> newPage
page <- newPage

### Actual Behavior
The page does not follow history.
page -> editPage (`history: false`)  -> newPage
page <- editPage(the page doesn't exist in history but is loaded)  <- newPage

### Additional context
In the fiddle you will also see the history debug with the `routeChanged` event

pablorodriguesnunes:
Please try to use `clearPreviousHistory: true` instead of `history: false`

Simone4e:
> Please try to use `clearPreviousHistory: true` instead of `history: false`

Exactly as the word "clear" says, instead of not entering the history page, you insert the page and delete the previous one, completely bugging the history

peitschie:
Interesting... I've generally used `reloadCurrent` when transitioning from the edit page => new page for achieving this effect.

E.g.,
```
<a href="edit/" data-reload-current="true">click to go edit page</a>
```

A quick dig, and it seems like when the previous page is kept in DOM, the router is ignored (which does feel like a bug to me!)

If you want a workaround, you can add `preloadPreviousPage: false` to your view initialisation, so the previous page is cleaned up. This will fix the routing!

```
var app = new Framework7({
  root: '#app',
  theme: 'auto',
  view: {
    // Ensures previous page get removed from the DOM
    preloadPreviousPage: false,
  },
```

peitschie:
Oh, p.s., that Fiddle is using v4 of the framework 😅 

Would suggest using these links instead:

* https://unpkg.com/framework7@7.0.8/framework7-bundle.js
* https://unpkg.com/framework7@7.0.8/framework7-bundle.css

Simone4e:
> Oh, p.s., that Fiddle is using v4 of the framework 😅
[...]

PR #4077, opened 2022-09-21
fix(core): Improve router back with history condition

Simone4e:
The `router.back ()` does not take into account the `history: false`, keeping the old page in the DOM, adding the` options.history` condition in the `navigate.js` file in the `keepOldPage` variable should solve the problem.

**Issue page:** [Link](https://github.com/framework7io/framework7/issues/4076)
**Actually code:**
```
let keepOldPage =
      (router.params.preloadPreviousPage || router.params[`${app.theme}SwipeBack`]) && !isMaster;
```
**Fix code:**

```
let keepOldPage =
      (router.params.preloadPreviousPage || router.params[`${app.theme}SwipeBack`]) && !isMaster && options.history;
```


peitschie:
The fix seems to be missing from here by the way @Simone4e ... looks like this is just trying to merge v8 into master 🙂

Simone4e:
@peitschie seems like I wrong something 😅,  can you open the issue for merge with framework? I will close that

PR #4117, opened 2022-12-26
Update notification.d.ts

jmarceli:
Notification `render` method is missing parameter type, which should be the notification itself according to https://github.com/framework7io/framework7/blob/master/src/core/components/notification/notification-class.js#L203


nolimits4web:
Merged, thanks!

PR #4127, opened 2023-02-02
fix(panel): fix panel close routine if it was closed immediately

### LINKED ISSUE #4108: Opening and closing panel quickly breaks any further tap/click events
- Framework7 version: 7.0.8
- Platform and Target: Chrome + Firefox browsers, Android Cordova
- Live Link or CodeSandbox: https://codesandbox.io/p/sandbox/dazzling-feather-0s7qc6

### Describe the bug

If a panel is closed very quickly after opening (can happen if the user accidentally double-taps, for example), the app suddenly becomes unresponsive to further tap/click events. It seems that the backdrop element is left behind, despite the panel being closed.

### To Reproduce

See CodeSandbox link: https://codesandbox.io/p/sandbox/dazzling-feather-0s7qc6
Tapping the "Open panel normally" shows a normal behaving panel. Tapping "Open panel and break app" will show no panel, but after tapping, no other link will work (i.e., app appears unresponsive).

In essence, the key is
1. Start the panel opening
2. Very quickly, close the panel (e.g., on the `panelOpen` event). This can be done manually by tapping very quickly on the panel open button, as it's possible to tap on the backdrop while the panel is still sliding open
3. When the bug happens, the panel is closed, but the app stops responding to any taps or clicks

### Expected behavior

The panel should close, but the app should remain responsive

### Actual Behavior

The panel closes but the app becomes unresponsive to taps

### Additional context

The root cause seems to be that the `transitionEnd` trigger never fires if the CSS properties are reverted quickly enough. Surprisingly, it's possible to visibly **see** the animation running when manually tapping quickly to cause the bug.

One possible idea (I'd be happy to raise a PR if you'd like) is to use `Element.getAnimations()` to see if the `transitionEnd` callback might not fire: https://developer.mozilla.org/en-US/docs/Web/API/Element/getAnimations

Basically, the panel open/close logic would be updated to set the CSS, then queue up a microtask (or even `setTimeout(..., 0)` task) which checks if `getAnimations` returns an active transition. If the check detects no transition and the `transitionEnd` callback hasn't fired, the callback would then be fired.

As pseudo-code, something would be added at https://github.com/framework7io/framework7/blob/94ffca087841e0b8e8c305e37157c0d9ddd0e3e0/src/core/components/panel/panel-class.js#L460
```
    function panelTransitionEnd() {
      transitionEndTarget.transitionEnd(...);
      // NEW: handle case where transitionEnd is not called
     queueMicroTask(() => {
       // TODO more accurately detect transition animations
       if (transitionEndTarget[0] && transitionEndTarget[0].getAnimations().length == 0) {
         transitionEndTarget.trigger('animationend'); // TODO is this the right way to trigger this?
       }
    }
    if (animate) {
      ...
      panelTransitionEnd();
      $el.removeClass('panel-out not-animated').addClass('panel-in');
      panel.onOpen();
    } else {
      ...
    }

```

peitschie:
fixes #4108 (again)

In practice, the time until the transition start is variable depending on the browser, amongst other things. Android + Chrome is between 27 => 44ms with a fast phone, but can take longer under many circumstances.

This removes the hardcoded time and instead looks for the transition start.

Depends on the new `transitionStart` method added to Dom7 in https://github.com/nolimits4web/dom7/pull/62

### DIFF
--- a/src/core/components/panel/panel-class.js
+++ b/src/core/components/panel/panel-class.js
@@ -289,7 +289,7 @@ class Panel extends Framework7Class {
   onOpen(modifyHtmlClasses = true) {
     const panel = this;
     // eslint-disable-next-line
-    panel._openTimeStamp = new Date().getTime();
+    panel._openTransitionStarted = false;
     const app = panel.app;
 
     panel.opened = true;
@@ -442,6 +442,13 @@ class Panel extends Framework7Class {
     }
     const transitionEndTarget = effect === 'reveal' ? $viewEl : $el;
 
+    function panelTransitionStart() {
+      transitionEndTarget.transitionStart(() => {
+        // eslint-disable-next-line
+        panel._openTransitionStarted = true;
+      });
+    }
+
     function panelTransitionEnd() {
       transitionEndTarget.transitionEnd((e) => {
         if ($(e.target).is(transitionEndTarget)) {
@@ -457,6 +464,7 @@ class Panel extends Framework7Class {
       if ($backdropEl) {
         $backdropEl.removeClass('not-animated');
       }
+      panelTransitionStart();
       panelTransitionEnd();
       $el.removeClass('panel-out not-animated').addClass('panel-in');
       panel.onOpen();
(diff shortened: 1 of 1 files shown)


peitschie:
Thanks for the quick review and merge @nolimits4web 

Just wanted to note, the Dom7 library in Framework7 core will need to be bumped to the new 4.0.5 release to ensure the new method is available 😅

nolimits4web:
Yes, it should be there in F7 7.1.4

PR #4157, opened 2023-04-03
[v8] Re-added strong outline and dividers to lists in KS for iOS theme

DAnn2012:
[v8] Re-added strong outline and dividers to lists in KS for iOS theme


nolimits4web:
Did you just add it everywhere for no reason? Strong, outline and dividers are context based, for sure these shouldn't be used in cards, fab lost demos, etc

PR #4182, opened 2023-05-31
Fix backdrop-filter blur (add prefixes)

100100101:
Backdrop-filter didn't work in Chrome modile (ios) or Safari mobile (ios).
Now works in Chrome modile (ios) in Safari mobile (ios).

I also tested on Android Chrome, but the effect does not work there.

I'll take a closer look a little later.

I also propose to make auxiliary classes for forced affixing of backdrop-filter
Something like
```less
.translucent-type-1 {
	-webkit-backdrop-filter: saturate(180%) blur(var(--f7-bars-translucent-blur));
  	-moz-backdrop-filter: saturate(180%) blur(var(--f7-bars-translucent-blur));
  	backdrop-filter: saturate(180%) blur(var(--f7-bars-translucent-blur));
}
.translucent-type-2 {
    -webkit-backdrop-filter: blur(15px);
    -moz-backdrop-filter: blur(15px);
    backdrop-filter: blur(15px);
}
...
```


nolimits4web:
Hi,

Thanks, but it doesn't require these prefixes. Plus prefixes are controlled by autoprefixer, and webkit prefix is already there  as you can see in Kitchen Sink:

screenshot

100100101:
I used tree shaking less files, like this:
Снимок экрана 2023-06-01 в 11 56 44
And not using postcss autoprefixer, and have this:

Now i add autoprefixer to my vite configuration:
Снимок экрана 2023-06-01 в 12 13 52
And now all working good, like full bundle (but with tree shaking ability) like this:

Perhaps you should mention this point in the documentation
Thank you)

PR #4221, opened 2023-09-08
fix(core): replace references to obsolete addListener/removeListener

peitschie:
As per https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList

Not a big deal... just produces some linting warnings for me in my resulting package 🙂

### DIFF
--- a/src/core/components/app/app-class.js
+++ b/src/core/components/app/app-class.js
@@ -198,8 +198,8 @@ class Framework7 extends Framework7Class {
     const app = this;
     const html = document.querySelector('html');
     if (app.mq.dark && app.mq.light) {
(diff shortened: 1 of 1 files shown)


peitschie:
Updated.. I didn't read the MDN article closely enough 😓 

> [addListener()](https://devdocs.io/dom/mediaquerylist/addlistener) Deprecated
>    Adds to the MediaQueryList a callback which is invoked whenever the media query status—whether or not the document matches the media queries in the list—changes. This method exists primarily for backward compatibility; if possible, you should instead use [addEventListener()](https://devdocs.io/dom/eventtarget/addeventlistener) to watch for the [change](https://devdocs.io/dom/mediaquerylist/change_event) event.

PR #4231, opened 2023-11-15
Bugfix: Fix there are types at 'framework7.d.ts' but this result could not be resolved when respecting package.json

### LINKED ISSUE #4181: There are types at `node_modules/framework7/framework7.d.ts` but this cannot be resolved when respecting package.json "exports"
- Framework7 version: `8.0.5`
- Platform and Target: `React with Capacitor`

### Describe the bug

Framework7 (core module) typings do not work when used in a typescript file. They do however work when used in javascript.

### To Reproduce

1. Create a new vite project using `npm init vite`. Select React with Typescript.
2. Install `framework7` and `framework7-react` with NPM
3. Import `framework7/lite` in `src/main.tsx`. The error should be visible. The framework7 library will not be typed
4. Rename `src/main.tsx` to `src/main.jsx`. The error is now gone, and the library is properly typed.

### Expected behavior

The library should be typed and the error should not be there

### Actual Behavior

There is an error while importing the types.

### Screenshots

The error in typescript file:
image

Renaming to js file fixes the issue:
image

### Additional context

None

anisjonischkeit:
Fixes #4181

I'm importing `framework7` in an astro project (vite) and I get the same error as in the above issue. 

> There are types at 'Documents/my-project/node_modules/framework7/framework7.d.ts', but this result could not be resolved when respecting package.json "exports". The 'framework7' library may need to update its package.json or typings.

The above seems to fix it. I know very little about this project so if this isn't the right approach (or it causes other issues) feel free to close.

Using: 
node 18
Typescript 5

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -4,12 +4,30 @@
   "description": "Full featured mobile HTML framework for building iOS & Android apps",
   "type": "module",
   "exports": {
-    ".": "./framework7.esm.js",
-    "./core": "./framework7.esm.js",
-    "./bundle": "./framework7-bundle.esm.js",
-    "./lite": "./framework7-lite.esm.js",
-    "./lite/bundle": "./framework7-lite-bundle.esm.js",
-    "./lite-bundle": "./framework7-lite-bundle.esm.js",
+    ".": {
+      "import": "./framework7.esm.js",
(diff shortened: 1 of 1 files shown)


peitschie:
Just confirming, this has fixed the issue for me as well using VueJS 🙂

PR #4235, opened 2023-11-30
Fix param type of calendar parameter formatValue

### LINKED ISSUE #4236: Type definition error: Calendar.Parameters.formatValue has a Date[] parameter
- Framework7 version: v5 and up

The typescript type definition is incorrect. The parameter type should be a Date array instead of a Date. The JSDoc comment also describes this.

### To Reproduce

Use typescript type checking.

### Screenshots

image

gewoonwoutje:
Change parameter type of values to an array, in line with the JSDoc comment.

Fixes #4236


nolimits4web:
Merged, thank you!

PR #4248, opened 2024-02-17
Fix sorting arrow direction

stanleyxu2005:
Sort arrow in data-table is wrong. When adding `sortable-asc`, the arrow is down, when adding `sortable-desc`, the arrow is up. 
The PR would fix the arrow direction issue.


nolimits4web:
> When adding sortable-asc, the arrow is down, when adding sortable-desc, the arrow is up.

This is the correct behavior

PR #4252, opened 2024-03-03
disable cache for device-info on server

antoniopresto:
fix  device info cache on server.


nolimits4web:
Thank you! I merged it manually with some tweak

PR #4305, opened 2025-01-19
fixes #4304 - popup.params sometimes undefined

### LINKED ISSUE #4304: undefined is not an object (evaluating 'i.params.swipeToClose')
In vanilla/core 

Seeing this error from iOS users - managed to track it down to specifically when users use the iOS app switcher/or the app loses focus whilst the popup is open - when user returns to the app the error is triggered:

`undefined is not an object (evaluating 'i.params.swipeToClose')
`

Have submit [PR](https://github.com/framework7io/framework7/pull/4305#issue-2797722217) to resolve this.

Image

scp:
In core/vanilla - seeing this error from iOS users - managed to track it down to specifically when users use the iOS app switcher/or the app loses focus whilst the popup is open - when user returns to the app the error is triggered:

undefined is not an object (evaluating 'i.params.swipeToClose')

framework7-ss

### DIFF
--- a/src/core/components/popup/popup-class.js
+++ b/src/core/components/popup/popup-class.js
@@ -184,7 +184,7 @@ class Popup extends Modal {
       $el.transition(0);
 
       if (
-        typeof popup.params.swipeToClose === 'string' &&
+        popup.params && typeof popup.params.swipeToClose === 'string' &&
         direction !== popup.params.swipeToClose
       ) {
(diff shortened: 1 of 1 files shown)


nolimits4web:
The question is why in your app they end up like this? In your console log it specifically says that popup is destroyed. I guess you shouldn't destroy it then if you still need it

PR #4307, opened 2025-02-02
feature(vue): support v-model binding on smart-select

peitschie:
This merge request adds a `v-model:smart-selection` model binding for the smart list element when used on an `f7-list-item`. Example syntax (as per the kitchen sink update) is something like:
```vue
 <f7-list-item
  v-model:smart-selection="predictedCoinFlip"
  title="Predicted outcome"
  smart-select
  :smart-select-params="{ openIn: 'sheet' }"
>
  <select name="coin-flip">
    <option value="heads">Heads</option>
    <option value="tails">Tails</option>
  </select>
</f7-list-item>
```

The key driver for this change is I wanted a more ergonomic way to programmatically control which item is selected in a smart-select component.

I don't particularly love the model name of `smart-selection`. Alternatives I considered:
1. extend the existing `value` prop to be a model (i.e., add a `update:value` emit)
2. call it `selection` or event `select` instead, but I felt this dropped the clear link to the smart-select component?

This is a completely random drive-by feature, so I'm absolutely fine with it being closed if it's not something you want. If you like the concept but want changes made, I'd be thrilled to adjust the patch to your liking 🙂 

Also, I'd be happy to update the docs as well if this feature is accepted.

Thanks for all your hard work on the framework!

### DIFF
--- a/kitchen-sink/vue/src/pages/smart-select.vue
+++ b/kitchen-sink/vue/src/pages/smart-select.vue
@@ -66,18 +66,84 @@
         </select>
       </f7-list-item>
     </f7-list>
+
+    <f7-block-title>Vue Model</f7-block-title>
+    <f7-list strong-ios outline-ios dividers-ios class="no-margin-top">
+      <f7-list-item group-title title="Coin flip guess" />
+      <f7-list-item
+        v-model:smart-selection="predictedCoinFlip"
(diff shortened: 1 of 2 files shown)


peitschie:
Ah good... I was about to ask @wxfred 🙂

PR #4400, opened 2026-08-16
fix incorrect toggle button style in RTL layouts

mayswind:
This PR fixes a display issue with toggle buttons in RTL layouts (#4401). The styles before and after the fix are shown in the table below.

| Before | After |
| --- | --- |
| image |  image |
| image | image |
| image | image |


nolimits4web:
Merged, thank you!

Issue #4402, opened 2026-08-16
Segmented button style is incorrect in RTL layouts


mayswind:
- Framework7 version: 9.1.2
- Platform and Target: iOS 26.6

### Describe the bug

In RTL layouts, segmented buttons are displayed incorrectly. The screenshots show the states when each of the three buttons is selected.

### To Reproduce

1. Import the RTL CSS
2. add an `f7-segmented` component with three `f7-button` components

### Expected behavior

When a segmented button is selected, the bold and highlighted states should appear together.

### Actual Behavior

See the below screenshots

### Screenshots

<img width="351" height="51" alt="Image" src="https://github.com/user-attachments/assets/cd548793-6a78-4c0b-a73d-a92dd8c4dc10" />  <br/>  
<img width="350" height="48" alt="Image" src="https://github.com/user-attachments/assets/e1b30814-d445-4d56-959f-e702da8c9ed5" />   <br/>  
<img width="352" height="48" alt="Image" src="https://github.com/user-attachments/assets/9013b32f-0104-40ca-bcae-35354e5876be" />

### Additional context

I have submitted a PR #4403 to fix this

PR #4403, opened 2026-08-16
fix incorrect segmented button style in RTL layouts

mayswind:
This PR fixes a display issue with segmented buttons in RTL layouts (#4402). The styles before and after the fix are shown in the table below.

| Before | After |
| --- | --- |
| Image | image |
| Image | image |
| Image | image |


nolimits4web:
Merged, thank you!

